PR #2, opened 2019-02-03
Small changes to get DFTFringe compile on Linux

hzeller:
Hi,
In order to compile DFTFringe on my Linux machine (running Debian sid), I had to do some slight modifications; this will hopefully others to get this platform-independent program running for them.

Mostly this was header filenames due to case-sensitivity on Unix. Also some macros looked like they were not in sync anymore (shomem); updated these.

There was a small change needed which probably is due to a newer version of opencv (cv::phaseCorrelateRes() does not exist anymore, this is now the same function call as cv::phaseCorrelate()).

Please have a look if it still compiles for you on window, but the changes should be mostly benign and work on both platforms.

Thanks for DFTFringe!

Changes needed below:
o some include-files were using a different casing than the files
  were named.
o Some opencv include files requires to be explicitly included.
o HUGE does not exist, using std::numeric_limits<> instead
o cv::phaseCorrelateRes() is now covered by cv::phaseCorrelate()
  (at least in current versions).
o Fix a few typos when doing bit-map matching: use & instead of &&
  operator (e.g. in event->modifiers() & Qt::AltModifier)
o Use 'const' qualifier for const* strings in code.
o Mention all the needed libraries in the qmake project file.


sellingrich:
so how do you get it to work?

hzeller:
With these changes, it compiles

Then, with
```
qmake && make
./DFTFringe
```

it runs.

ker0sin:
Hey, it compiles, but doesn't unwrap phase.
It's possible to fix it by using std::numeric_limits<double>::lowest() instead of std::numeric_limits<double>::min() in punwrap.cpp and dftarea.cpp.

hzeller:
@ker0sin thanks for the suggestion, fixed (I am still in the grinding phase of my mirror, so I have not actually used DFTFringe yet on Linux :)

@githubdoe ready to review and merge.

gr5:
I'm sorry this pull request was "ignored" for so long.  Pull requests are now being paid much more attention to.  The latest version of DFTFringe compiles nicely in linux.  Changes are still happening.  I find your comments on this PR helpful and may refer back to them in the future.

Going forward, please feel free to submit more pull requests as they are being paid more attention to.

Issue #6, opened 2022-11-20
"show resized image" crashes application


atsju:
I load wat.jpg, autotrace outline, done, then "show resized image" crashed the application

version 6.0 on win10

githubdoe:
This is not the place to talk about problems you may have with the program.  Instead use the groups.io interferometry site.  At
https://groups.io/g/Interferometry.   I will look for messages from you there.

atsju:
I'm firmware engineer and "in general", open source software take problems (I understand for the part "no question") on the issue tab in GitHub. If you want not issue here, maybe you should go to "settings" on github repo and untick the "issues" feature.

githubdoe:
Posting issues here is fine except I don't pay attention to it and find it easier to discusses on the other site.

atsju:
Not sure to fully understand, these are quite opposite messages. Anyway, I will post on groups.io if you prefer so. 

Some last "in general" or "industry standard" (please take no offense, I just do this professionally and also helped on some small open source repos like yours. The tool is great!): GH issues are closed when the code is updated with a fix, even better if the issue is linked to a pull request fixing the actual code. Pull requests and commit with small changes (with peer code review) are really recommended. Versions number for releases should follow https://semver.org/

PR #7, opened 2023-02-02
linux compile

brave-ulysses:


atsju:
Hi @brave-ulysses 
I don't know if you are aware, but only 1 person is maintaining this repo and he doesn't use PR....
For the full story I would recommend you read this https://groups.io/g/Interferometry/topic/95167900#32343 and maybe try to defend the github story yourself. Also forget issues (see https://github.com/githubdoe/DFTFringe/issues/6) as everything here happens on groups.io ...
PRs are not even looked at ... (see https://github.com/githubdoe/DFTFringe/pull/5 which has neither been merged nor copied manually during last update).

I just wanted you to know so you are not too disapointed.

githubdoe:
Now that I understand what a pull request is and what to do they are being processed.  In this case by a collaborator.  Brave-U is also a collaborator and we 3 have been in touch.

atsju:
Good news :)  thanks for update.

gr5:
I'm in the process of doing the merge but I need to be able to compile in linux first so it could be a few days.

After this PR is complete, I'm happy to look at the other's.  By the way, Dale gave me "collaborator" status for this repository.

peytondmurray:
Hi all, I'm also interested in seeing support for other platforms on the project. At the risk of hijacking the discussion here, I'd be happy to make a PR that uses meson as a build system so that we don't have to worry about setting the correct include directories for dependencies in the `DFTFringe.pro`, etc. This is something people would be interested in?

On a related note, @gr5 how do you feel about outsiders making PR reviews here? I'd be interested in giving input on this PR but I wanted to check with you before making an unsolicited review.

gr5:
It's okay to hijack the discussion on this PR thread.  Wow - 3 people want to make DFTFringe work on more platforms!  That's fantastic.  So I committed and pushed the "linux compatible merge" so you can look at it now.  It's branch "gr" and hope that Dale will merge it into master soon or I can do it if he wants.

I'm not familiar with meson but it sounds like a good idea to me.

So start with branch gr at this moment.  Which should change to branch master eventually.  We might want to tag branches that have been successfully merged to linux with some kind of "linux" tag.  Hmm.  Can tags move?  Google says "yes".

brave_ulysses has created a tiny little dockerfile that can build DFTFringe for ubuntu.  I modified it to build for ubuntu:focal (the version of ubuntu I have) and it is getting one build error as of now.  If you are interested in the dockerfile, let me know.  Even if you don't want to use docker, if you are using any form of linux, the dockerfile has all the "apt install" commands that you will need.  It is not checked into the repository at this time. The steps in general are: 
1) apt install a ton of libraries including qt creator 5.
2) get the source code from github
3) do: qmake
4) do: make

DFTFringe.pro
```diff
@@ -96,7 +96,7 @@ SOURCES += main.cpp\
     squareimage.cpp \
     bathastigdlg.cpp \
     zernikeeditdlg.cpp \
-    settingsgeneral2.cpp \
+    settingsGeneral2.cpp \
```
peytondmurray, line 99: Can we stick to using lowercase names here? There's already something of a convention for this. Windows by default is case insensitive, but linux is by default not. It would be good to assume case sensitivity here for compatibility.

peytondmurray:
@brave-ulysses I'd also request removing the build artefacts from this PR; they're platform specific and there's no guarantee they'll work on everyone's system. If @githubdoe wants to do releases binaries can be built then.

moc_settingsgeneral2.cpp
```diff
@@ -0,0 +1,227 @@
+/****************************************************************************
```
peytondmurray, line 1: See earlier comment about removing build artefacts.

punwrap.cpp
```diff
@@ -27,6 +27,8 @@ double* qmap = NULL;
 double* phase;
 double* unwrapped;
 
+double _HUGE=1e37;
```
peytondmurray, line 30: Is this an arbitrary large value? Why not use [`DBL_MAX`](https://en.cppreference.com/w/cpp/types/numeric_limits/max)?

githubdoe:
The upper lowercase for file names comes about due to Qt.  When I make a new Qt class it creates the file names from that class according to it's case preferences.  I'm not going to change that behavior.

Dale

githubdoe:
Years ago I was able to use Huge as a huge number when needed.
Then Qt lost that ability.  That was my fix at the time or the author of punwrap's fix.  No longer remember.  Too many different package have various ways to specify or not Huge numbers that I have not kept up with and don't intend to change.  Never heard of DBL_MAX.  You have to realize that I'm stuck mostly back in the world of original C++ and know very little of what ever is the latest spec.

githubdoe:
I had perhaps a false memory of how Qt created files names.  It might only be based on how I named the class.  In any case I am surprised by the settingsG...    file instead of the settingsg...   

I'm also worried that the linux may not be using the same OpenCV version that I did.  I don't know if that causes problems but don't have time to figure it out.

peytondmurray:
> I had perhaps a false memory of how Qt created files names. It might only be based on how I named the class. In any case I am surprised by the settingsG... file instead of the settingsg...

Agreed - I'd definitely prefer to fix the case-insensitive import statements rather than change the name of a bunch of files.

> I'm also worried that the linux may not be using the same OpenCV version that I did. I don't know if that causes problems but don't have time to figure it out.

I have no idea what OpenCV version @brave-ulysses was using - this is one of the reasons it would be nice to have a platform-independent build system in place, so that we can specify dependency versions without manually having to edit the includes for the project. I'm in the middle of incorporating meson to build the project, and I'm hoping that will help with this.

gr5:
You are looking at the wrong version.  Get the latest "gr" version.  Did I not do a push?

HUGE is gone.  It's now HUGE_VAL.  I think it comes from math.h.

brave_ulysses had to mix up the case to get the include statements to work because the filenames were like that unfortunately.  We could have a commit that just makes all files lower case at some point.  But you have to change the files and also change the includes at the same time and windows would be confused as it doesn't differentiate.

the moc_ files are gone.  You have the wrong version from git.  Get the latest "gr" version.

Makefile is also gone.  It's created by the qmake command.

gr5:
I *must* have done a push because I pulled it over to linux and built it there and it wouldn't have built with that "HUGE" stuff.

gr5:
I think I know what happened - you are looking at each of the changes one at a time - one commit created the HUGE 1E37 code and checked in a few moc_* files (brave_ulysses) and then a later version removes them again.  Better to do something like "git diff master..gr"

gr5:
So if anyone gives me their emails I'll send you the Dockerfile that has versions of opencv, amadillo, etc that will work as they are tested.  Dale isn't using the latest bleeding edge code so possibly a version from a year or two ago might be fine.

brave-ulysses:
using opencv 4.5 on linux mint 21.1

brave-ulysses:
for platform independence, i'd vote python...

suboptimal numba and cuda code processing 12+ 2048x2048 ffts per second
https://groups.io/g/Interferometry/message/32330

githubdoe:
Unless more has changed in the last 7 years since I tried a version in python it was about 100 times too slow.  Python is on of my preferred languages.  7 years ago I first started DFTFringe using python with wxWidgets.  But it's math on large matrixes is way too slow. 

If anyone want go ahead and try implementing a python version.  But not in this repro of course.

githubdoe:
Also one thing about python is usually you have to have the user install python and all the libraries the program needs along with it.  That is way to much to expect many ATM's using windows to do.

gr5:
rewriting DFTFringe from scratch in python seems like a lot of work.  Or did you just mean use python to control the build process?

I suppose you could leave out 70% of the features as many are not critical features.  Just do the most basic stuff.  Anyway I'm not sure which step is the slowest step but it could be the "best fit" step where it does the best fit for the first 47 or so zernike shapes.  On the other hand you don't really need to do 47 zernike's.  Just doing a few would be quite useful (tilt, defocus, astig, S.A., coma).  Or maybe the unwrap step is the slowest?  I really don't know.

githubdoe:
There are several compute bound steps.  Each FFT, unwraping, using the Vortex routine. Computing Zernike polys as well.  They depend upon the spatial resolution needed.  Well of course all routines do.

If one were to convert to Python it must have the same functionality as the current version.  Otherwise it becomes DFTFringe lite.  There are reasons for each of the functions.  Not all are used by each user but all are used.

moc_settingsgeneral2.cpp
atsju: Maybe a .gitignore file should be created

DFTFringe.pro
atsju: I agree, use actual file name

punwrap.cpp
atsju, line 30: I would add, if a variable is meant to never be modified you should write
```suggestion
const double _HUGE=1e37;
```
even better with C++11 use constexpr
```suggestion
constexpr double _HUGE=1e37;
```
I heavily recommend switching to C++11 for many of it's good new features. Static asserts, constexpr , nullptr, list initialisation, range-based for loops and ranges in general.
https://en.cppreference.com/w/cpp/11
It's not just fancy new words, it's a paradigm shift and a better safer language

atsju (commented):
Please share me the docker file. I might do a PR to add github actions for continuous integration.

I don't know if you noticed but github tells you "This branch has conflicts that must be resolved".
Your branch is 1 commit ahead and 8 commits behind main repo. If you need help with git or github you can contact me. Also you should work in a branch different of "master" in your fork. Now that you have maintainer rights, you should be able to work in original repo directly on a dedicated branch instead of a fork. It will be more easy that maintaing the fork up to date if you are not used to git (master shall be protected with branch protection rule so that nobody pushes there by error without PR, even githubdoe)

punwrap.cpp
gr5: I know git quite well and am very comfortable with git.  However I am _not_ familiar with pull requests and github.  I mean I've used github quite a bit but I've only ever created one PR and it was just pulled with no fanfare.

I don't have a fork of DFTF.  I am working on it directly.  The changes for this pull request are in branch "gr".  I want to merge the changes back to master.  But I want Dale to agree.  What is the normal procedure for handling this PR.  Do I somehow tell this PR that it should be looking at branch gr?  Do I merge them back to master and *then* will github automatically realize that the PR has been merged?  I should go read about it but since you seem to know more, please let me know.  I could add the const.  Adding .gitignore is great.

Adding Makefile or moc_* files was never a problem for Dale as qt creator puts the generated files in a different folder.  But for linux users a .gitignore would be very helpful.  I'm happy to do those things but again, you are showing suggested changes on the wrong branch.  The "const" doesn't apply anymore, I believe.  That line of code is gone I believe.  90% sure.

How do I tell this PR to point to the latest gr branch instead of the brave-ulysses commit?

atsju: @gr5  this would be a 5min talk on discord if you are interested (or 1h if we like the discussion)
I opened #8 for you with correct branch. The PR we are on, is brave_ulys forked and modified master branch.
image

Maybe you can modify the PR, easyest is just to close and look at #8 only

Dale does not know PR and GitHub eather so I would recommend I help you on the standard "process" on open source. It might sound like I want it my way but I honestly think it's your best option. I worked on many repositories and even if you think I do lack the DFT fringe experience I will never harm your code without permission. However I like to interfer in the process thing.

gr5:
Closing this PR so please, if you have comments specific to the PR or related to building in Linux in general let's move over to PR #8

PR #9, opened 2023-02-12
delete autogenerated files from repo

atsju:
This folder has been added in v1.6 but all files are autogenerated ans shall not be added in the repo.

I tested the build on linux through windows WSL to confirm the files are not needed.

PS: I will add a .gitignore file to avoid such errors in a different PR once #8 is merged


gr5:
Oops.  Yes, these should not be part of the repository.  There is no harm but lets get rid of unneeded junk

gr5 (approved):

githubdoe (approved):

PR #11, opened 2023-02-17
bump OpenCV from v3 to v4 on windows to have same build as linux

atsju:
Hi @githubdoe 

After all these month I have finally been able to build on windows locally.
I used latest version of everything I found and many discrepancies between linux and windows build were only due to different version of openCV. I think it simplifies much if everybody uses openCV v4. 
Many deprecation warnings are here during build, I don't know if this was already on your build, anyway it can be fixed later in different PR.

I tried to change the `.pro` file minimally for you so I have not been able to test _this exact_ file.
I you think the idea is good, could you please try this build with latest openCV on your side ?

In the meantime I will work on CI/CD for Windows to avoid this work in futur.

PR #12, opened 2023-02-17
Jst/windows ci

atsju:

PR #10, opened 2023-02-15
add continuous integration (CI) for linux build

atsju:
This will automatically lead to a build on each PR for linux. Thanks @gr5 for having shared your dockerfile for building.
If build fails, we know we must not merge.
See github actions documentation to learn more. You have 2000m free per month on server for automated build; After that either you pay or it does not build anymore; But the free plan should be more than enough here.

I also added a .gitignore file to avoid adding unneeded files by error

I will add windows build in different PR because it's much more complex.
And the maybe continuous deployment (CD) to male release from a simple tag. But this is not a promise.

PR #13, opened 2023-02-18
remove 19 #ifndef statements and simplify code

gr5:
This PR removes 19 unneccessary #ifndef statements.  4 were related to beep() and sleep().  15 were related to code added by @brave-ulysses .

I'm embarrassed.  When I saw the PR from @brave-ulysses - he changed a ton of include statements for opencv that were quite a bit different from the windows version.  I assumed they were **incompatible** because they were somewhat different versions of opencv (I assumed).  So I added about 15 #ifdef statements assuming they were needed.  What I didn't realize is that the newer include statements from @brave-ulysses work fine in windows.  So no #ifdef were needed.

This PR works fine in both windows and on my ubuntu.  I did quite a few tests - ran DFTFringe through the basic features but not 100% of them.  But all I'm doing is changing include statements.

I did also check the beep() feature.  It works both in linux and windows.  So now we are using qt5 functions designed to work cross platform to implement beep() and sleep().


atsju:
This will clash (merge conflict) with my #11. I think This one can be merged and we close #11.
Next time you could cherry pick my commit and work on it so we see each other modification clearly (you still could and force push if you feel at ease with git).

brave-ulysses:
hi george,
iirc, either the c or the c++ include files could have been selected and the corresponding function call changed to suit; i can't say for sure that i did it in a consistent manner. ie, some used c++ include and function changed to cv::<f_name>, some used c include and cv.<f_name>

might pay to take a look at this...

githubdoe:
My only concern is how OpenCV gets compiled and the DLL's joined with this archive.  It must be compiled with the same compiler and sometime the same version of the compiler as DFTFringe is compiled with.

gr5:
> My only concern is how OpenCV gets compiled and the DLL's joined with this archive. It must be compiled with the same compiler and sometime the same version of the compiler as DFTFringe is compiled with.

Okay.  I'll check.  

**win**

So on windows I'm using the exact dlls Dale compiled as I took it from his installer.  it appears to be opencv 3.4.12 as all the dlls have "3412" on them.


The compiler I use on windows, when I do "help about" mentions "qt creator 9.0.0" "based on QT 6.4.1".  qmake version is **5.15.2**.  It works fine.  I've built dftf many times and everything seems to work great.  I've used many different parts of DFTF.

On **linux**, the things that get installed:
For ubuntu:focal it's "qt5" where qmake is "5.12.8"
opencv-core4.5d

For ubunto:latest (the autobuild on github now) it is also "qt5" where qmake is "5.15.3"
opencv-core4.2

It all sounds like a mess but I assure you Dale it builds and runs in linux very nicely and the windows version builds and runs on my windows machine very nicely.  I've tested many functions and features in both the linux and windows versions.  The only differences I've noticed so far are minor and have to do with font sizes.  The opencv libraries work great.

atsju:
> So on windows I'm using the exact dlls Dale compiled as I took it from his installer. it appears to be opencv 3.4.12 as all the dlls have "3412" on them.

I don't think it's OK to use 3.x DLL and point to 4.x include files. This is why in #11 I also had to change the .pro file. I rebuilt everything with same compiler.

Dale, if you have any concern about if build works or not, maybe you can wait a bit, I'm still on the process of making github so the Windows build. it's not easy and might take weeks but I'm confident. At the end you will even be able to download the archive and test without building. Build will be done one clean machine each time so no cheating.

Also I know you don't want to install inux but are you on windows 10 ? If so, you can just use WSL (run `wsl --install` once) and build in a linux console without installing linux. It works great and application even opens the GUI in windows when launched from the WSL command line

githubdoe:
Very few OPencV functions signature have change overtime.  IIRC some have but that was several years ago.  So that makes mixing modern headers with old .dll more or less ok.    But there is a risk.  The complication is the there are standard build version of OpenCV for windows machine but those were built with Visual Studio Compiler and are not compatible with DFTFringe.   That is ok because the proper .dll files are delivered with the install.  The windows install is more or less hand crafted by me.  

Just to let others know the process.  To start with I move all the Qt, OpencCV, Qwt and so on .dlls into the directory with the DFTFringe.exe.  I might even remove them from the windows path.  Then I run DFTFringe and add any .dlls it announces it needs.    At this point I have more dlls than needed.  To find the ones not needed I try to delete all the dll files in the directory while DFTFringe is running.  Only the ones it needs can not be deleted the others are deleted.  I only have to do that once for a major release like when going from 32 bit to 64 bit versions.

gr5:
> I don't think it's OK to use 3.x DLL and point to 4.x include files.

I don't think anyone is doing that.  The include files I point to on my windows build machine are from 3.4.12.  As are the dlls I link against.

I don't completely understand why there are different sets of include files but @brave-ulysses implied that there are ".c" include files and also ".cpp" include files and either set of include files will work with Dale's code.  But then why won't either set work with the linux code?  I don't know.

atsju:
from my understanding, the IFDEF/ELSE where due to folder architecture of openCV3 VS openCV4. 
But apparently I was partially wrong because include/opencv2 exists for a long time (at least 2.2 in 2010). However include/opencv folder disappeared with 4.0.0 https://github.com/opencv/opencv/tree/4.0.0/include

Maybe Dale was using openCV3 with folder include/opencv and @brave-ulysses did only do minimal changes but used openCV4 and this is how he made it work. I think this is only an artifact from wanting to do minimal changes over time. We should just try build with openCV4 and if it work then it works. Simplify the code and all good.

gr5:
Thank you for figuring this out and explaining it.  Yes, that makes the most sense.  Dale using opencv3, brave_ulysses using opencv4.  opencv2 includes still work in both.  opencv includes don't work anymore in opencv4.

gr5:
So I think this PR makes more sense than ever.  It moves the includes to opencv2 style includes (even though Dale uses opencv3 and linux builds are using opencv4).

atsju:
Yes this is exactly what I think. Please have a look at #11 because I also updated the `.pro` file.

Also @githubdoe should use opencv 4.x.y locally if not already the case.

gr5:
> Please have a look at #11 because I also updated the `.pro` file to get correct/better version 4 of openCV DLL.

PR11 is the same as this pull in that you make all the include changes exactly the same as I did and:
- #10 also removes ifdef's around beep() and sleep()
- #11 has .pro changes to switch opencv to 4.6.0, qwt from 6.1.5 to 6.1.6,

atsju:
it's #12 that removes ifdef's around beep() and sleep() not #10
You make the changes as I did, not the other way ;) 
Anyway yes #12 is better, I let you close #11 when this is merged. I think I misunderstood, I though you were also building openCV 4 and not updating .pro. If you build with openCV 3.4.12 then it's fine and can be merged as is. 

We can bump the openCV version later.

githubdoe:
I hate building OpenCV.  It took me many tries over several months to get one that even built and was compatible with DFTFringe build.  Thus I avoid building it if at all possible.  That is the reason I stayed at the version I did.

atsju:
I have no build failure with openCV 4.6.0 but there are many warnings for deprecation, so I suppose if we move, the code will need an update. Are you ready to make the move or should we understand you prefer staying on current openCV version ?

If we agree on mingw and Qt version I could "easily" make artifacts (zip with necessary DLL and include files built with same tools as you) with github actions that you can download :innocent:

githubdoe:
atsju,
I'm not inclined to move.  For me that is unnecessary work.  If we were to rework DFTFringe every time one of the libraries updates we would never get done.  For instance Qt is beyond the version I use.   One time they removed OpenGL version support for what I used.  It took a long time to upgrade.   However if someone wants to tack it go ahead.  But what I deliver in the install has to work for windows.

gr5:
It sounds like Dale doesn't want to update any libraries nor QT so let's pull this PR (#13) and close the others for now.

And maybe concentrate on actual bug fixes, or new features.  Features and bug fixes that Dale likes.  Maybe through the "issues" feature of github.

I built this PR on windows using the dll versions that Dale uses and it should be the same includes as well (but Dale please build this to be sure).  Dale actually expressed concern about the number of ifdefs.  So I delivered this simplification of the code.  I know Julien also did the same thing but didn't compile against Dale's exact dlls.

atsju:
Now that #10  has been merged it will be automatic but just for this time, if you close and reopen this PR, the CI should launch.

atsju:
Now that #10  has been merged it will be automatic but just for this time, if you close and reopen this PR, the CI should launch.

githubdoe (approved):

PR #14, opened 2023-02-21
Fix missing icon in windows build

atsju:
@githubdoe  I dont know how you build manually on your computer but without copying icon.ico or other changes it does not work on a clean build.

This fixes windows build. Note I don't know why it's not needed by Linux. Also we will see if it breaks Linux build thanks to CI


gr5:
Same thing in windows and linux for me - it builds without the icon but then you don't have the icon in the corner or when you windows-tab around to other applications - it doesn't have Dale's iconic icon.

gr5 (approved):

PR #18, opened 2023-03-17
modified "about" window to get compiler, QT, openCV and armadillo version automatically

atsju:
It's not perfect as armadillo version is got dynamically at runtime while other versions are static at compile time.

Next step could be to add windows/linux version from computer it's running on.
Also adding the sha of git commit where build comes from would be great to track issues (I know very few people use their own build but this might change a bit with Linux support)

On linux: 
image

With the windows build from github actions
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # DFTFringe
 DFTFringe Telescope Mirror interferometry analysis Program.
 This is the successor to OpenFringe which is a Telescope Mirror interferometry analysis program.
-This code depends on QT, OpenCV, and qwtPlot.
+This code depends on QT, OpenCV, QWT, Armadillo, Lapack and BLAS.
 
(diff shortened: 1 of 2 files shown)


atsju:
@gr5 I did a merge commit to avoid dismissing a potential review in progress but if you prefer I can rebase the branch to keep more linear commit history

gr5:
> I did a merge commit to

Great.  Actually I can see the build is in progress now.  I will build this on my own machine today just to double check everything and hope to merge this today.  I'm in the middle of a bug fix right now so I'll commit that to my own personal branch first (or do a git push) but I have errands this morning before that.

gr5 (approved):
Looks good.  As an aside - note that I'm pretty sure nothing armadillo related is used (no functions, no dlls) except for an .h file that defines a few variable types.  No actual armadillo functions are called.  So maybe we could "fix" that situation some time.

PR #15, opened 2023-02-22
Windows CI build

atsju:
I'm proud to announce this implements CI to build Windows version of DFTFringe. 
First build will take over 1h because of OpenCV build but every next build will use cache I setup and take under 10 minutes. Cache lives for 7 days.

You will notice I moved openCV 3.4.12 to 4.6.0. I think it is good practice to keep up to date to limit work and help maintenance. And I have not been able to build 3.4.12 anyway so there is no real question here. 4.7.0 breaks build but I might fix later.
I also made relative path for all dependencies in .pro. Maybe we have an other solution but this was easiest for me to make the CI work. Sorry Dale, you will either need to accommodate the path I propose or we set so env var or something to accommodate everybody. I'm open to discussion (chat or call).

I don't know why lapack and blas are .a instead .dll on my build, maybe we can fix that later.

You will be able to download a zip of build result exe and dll in action tab under artifacts. It's named `DFTFringe-build-artifact`. here is an example of the build at the moment of this PR: https://github.com/atsju/DFTFringe/actions/runs/4246078248
Once mergged, actions will run in DFTFringe organisation instead my atsju organisation.
Artifacts live for 90 days. It does not replace releases but will help Dale and others test the proposals without rebuilding himself.

@githubdoe If you write a simple but complete document about creating the installer I could add the step to the build so that the artifact is an actual installer instead the exe that won't work on it's own on a clean computer.

I keep this PR draft for now because I would like the other PRs to be merged before so I can rebase the branch properly.
I also expect to have some discussion here. 

This has already been a lot of work but I have a some improvements in mind. Maybe I should list them is an issue ?

Notify @githubdoe @gr5 @brave-ulysses

PR #27, opened 2023-05-30
fix linux build

atsju:
I was missing `apt update`  
see https://www.reddit.com/r/linuxquestions/comments/13vx7p0/failed_apt_install_intelmediavadriver/  
and https://github.com/actions/runner-images/issues/2924


atsju:
@gr5 please merge this one before #15 (not mandatory though if branch protection is not set or if you can override it)

gr5 (approved):

PR #28, opened 2023-05-31
Add issue templates

atsju:
I created some simple issue templates as explained here
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

I only slightly modified the proposed templates. This has no impact on code, only on GitHub usage.


gr5 (approved):
I'm not sure how useful this is but I don't have a problem with this either.  We certainly don't have an abundance of issues to be triaged.  Also it's hard to see exactly what it will look like.  But I'll approve this for now and then test it out.

PR #32, opened 2023-06-25
artifact with installer for windows

atsju:
In this PR I will use QT installer framework to make installers. I use this instead of historial NSIS because it is multi-platform.
good link to learn : https://www.youtube.com/watch?v=1pKMcwJZay4

### Before merge I need:
- [x] installer for windows
- [x] use windeployqt to list and copy dependencies instead manual work
- [x] fix #26 

### do this separate to keep small steps
- [ ] installer for linux
- [ ] check which licences would need to be prompted (I suppose none is required for end user. Double check is required) or added in installation folder

### This can be done in a separate PR as artifacts already benefit from installer even without release:
- [ ] action to make release in one click (no local build)
- [ ] automated version and date update in releases

### Mac is not supported today. Keep this for later
- [ ] missing build and installer for mac

PR #38, opened 2023-07-02
Releases are automatically drafted from semver tags

atsju:
### Clearly indentification of the running code:
Following #32, this PR make a a better artifact generation process. Now the artifact has clear ways to be identified. If somebody uses it (end user should only use tagged versions) they can know exactly the version they are installing and/or running.

### (semi-)automated release creation:
This enforces semver convention https://semver.org/ which is industry standard. Next release should be v7.0.0
7.0.0 will not create a release. Neither will 6.3 and v6.3. You need vx.y.z

The process for creating a new release is highly simplified. 

1. push a new tag respecting senver format (`git tag v7.0.0` `git push --tags`)
2. wait for workflow to finish. You now have a DRAFT release available on release page (everything has been generated and uploaded automatically)
image
3. re-test the installer to check everything is fine
4. change the text in release at your convenance (list of changes are automatically generated so you forget nothing)
5. publish the release

### here are some example of what we see:
PR build generates : DFTFringeInstaller_60c45320dcb2a2a792da264ef6b7b797ce3eee59_c18a74e0f4f4db36eddc66a68eb08e91cd6c6a1c.exe
image

Merge to master or push to master or manual run will generate:
DFTFringeInstaller_baa81492b15f4ab6ddf39689cb75cfac7a99a888.exe 
image

Pushing a tag will generate:
DFTFringeInstaller_v7.1.2.exe
image

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -5,6 +5,7 @@ on:
       - master
   pull_request:
   workflow_dispatch:
+  workflow_call:
 
 jobs:
   build-linux:
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -295,6 +295,27 @@ jobs:
         with:
           path: 'DFTFringe'
 
+      # Update the 2 strings automatically
+      # MY_AUTOMATED_VERSION_STRING
+      # MY_AUTOMATED_DATE_STRING
+      # TODO must work on tags but also on PR and manual builds (using github.sha ?)
+      - name: Find and Replace MY_AUTOMATED_VERSION_STRING
+        run: |
+          echo "${{github.event_name}}"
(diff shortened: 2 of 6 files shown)


atsju:
@githubdoe  @gr5  the semi-automatic release process is ready. I described everything above. Please have a look and let me know.

Just to be 200% clear, you can still get the installer to build automatically or manually without tag. But a tag with correct name will directly create a release draft to simplify your life. 
In the futur, it could also upload the linux and macOS things.

gr5 (approved):
I'm going to merge this.  I spent some time looking at the changes.  I don't understand it 100% but it seems like a good idea.  I've never used the "release" features on github so I don't quite get how you publish a release but this all sounds good.  We can always revert this change if there is a problem with it.

PR #40, opened 2023-07-04
add probem matcher to show build warnings in PR

atsju:
Hi guys, there is a crazy amount of build warnings. Maybe they are not a big problem but having so much warnings will make new (and potentially problematic) warnings invisible. GitHub has a [feature](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) to show the build warnings in PR. There is a limited number, so I need to fix current warnings so that it becomes really interesting. However this enables the feature.

It will show like 
image

For information: 608 warnings on windows build:
- 12 on ui files
- 52 Wunused-parameter
- 341 Wdeprecated-declarations
- 32 Wunused-local-typedefs
- 9 Wwrite-strings
- 52 Wunused-parameter
- 22 Wsign-compare
- 23 Wunused-but-set-variable
- 48 Wreorder
- 1 Wcatch-value
- 40 Wunused-variable
- 1 Woverflow
- 2 Wmisleading-indentation
- 12 Wmaybe-uninitialized
- 2 Wreturn-type

I might have missed some. And Linux build exhibits some different warnings that not all overlap with windows build.

Altough `misleading-indentation` is almost pedantic,  `overflow` is just lost in the mass and could have consequences.

### DIFF
--- a/.github/matcher/uic_matcher.json
+++ b/.github/matcher/uic_matcher.json
@@ -1,0 +1,16 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "uic-problem-matcher",
+            "pattern": [
+                {
+                    // match things like : generatetargetdlg.ui: Warning: The name 'layoutWidget' (QWidget) is already in use, defaulting to 'layoutWidget1'.
+                    "regexp": "^(.*\\.ui): (Warning):(.*)$",
+                    "file": 1,
+                    "severity": 2,
+                    "message": 3
+                }
+            ]
+        }
+    ]
+}
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -15,5 +15,7 @@ jobs:
       - run: sudo apt install -y apt-utils build-essential wget qt5-qmake qt5-qmake-bin qt5-assistant qtbase5-dev qtmultimedia5-dev libqt5charts5 libqt5charts5-dev libqt5multimedia* libqt5datavisualization5-dev libqt5datavisualization5 libopencv-core-dev libopencv-core4.5d libopencv-dev libqwt-qt5-6 libqwt-qt5-dev libarmadillo-dev libarmadillo10
       - run: qmake
       - uses: ammaraskar/gcc-problem-matcher@master
+      - run: echo "::add-matcher::.github/matcher/uic_matcher.json"
       - run: make -j4
+      - run: echo "::remove-matcher owner=uic-problem-matcher::"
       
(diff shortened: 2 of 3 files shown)


githubdoe:
As you can tell I long since gave up on trying to get warning free builds.  Feel free to address any that you choose.  Ui files are Qt generated so impossible to fix.

atsju:
image

More seriously, it was already my plan to fix as many as possible.

atsju:
> What is a problem matcher? What does it do? What is it for? Why would it be helpful for fixing warnings?

It's the feature that makes the annotations possible in PR. It searches for some specific regex in build log to extract the warnings. The build log shows colored lines with the matches (i.e. highlights warnings) and then the 10 first are shown in the PR. 

> Are most of the warnings in the ui files? I ask because it looks like you only are excluding those.

No. The number of warnings per source is shown in PR description. I'm not excluding Ui warnings but showing them. You don't see the CPP matcher file because it's just the line `using: xxxxxxx problem matcher` so I did not need to write the regex. So basically I parse for Ui + GCC warnings.   
Now, the caveat is that only first 10 warnings are shown in PR. This is why we only Ui files war into are shown in the PR and not all 600.

> I read the description on github about problem matchers - it creates annotations. But what are annotations? lol.

If you look in file change tab if the PR, you will see some comments that are added automagically. This are the annotations. It's better for the GCC warnings than Ui as it's on the exact line of the warning. Maybe I will remove the Ui matcher so that it highlights actual usefull 10 GCC warnings and not Ui ones. 

> I mean when I compile myself I get to see all these warnings (unfortunately). Why do I need github to also annotate them?

Because you don't build on Linux, neither on Dame versions of everything and some warnings do not overlap so you don't get to see them. Because when someone proposes changes, you don't need to rebuild yourself so you will see in the interface if they created warnings instead of letting them through. Of course this all makes much more sense in a zero warning policy.   
My next PR will probably be about solving all warnings. I'm not afraid about this work and after years of development I'm convinced there is always (99% of the time, maybe exclude embedded development and other specific fields) a way to make code actually warning free.

gr5 (approved):

gr5:
Okay.  Approved by me.  Is there any reason not to merge?  Last minute changes?

atsju:
Dale thinks I cannot fix the Ui file. Maybe I can remove the Ui regex for now to prioritize on GCC warnings first.

gr5:
Lets do that.  Remove ui files.  We can add them back in if someone figures out how to remove the warnings.  I'm just going to guess here that they use one or two deprecated functions and only QT people could update how ui files are converted such that they use newer functions.

atsju:
> Lets do that. Remove ui files. We can add them back in if someone figures out how to remove the warnings. I'm just going to guess here that they use one or two deprecated functions and only QT people could update how ui files are converted such that they use newer functions.

done. Now we have the 10 first GCC warnings only, on correct file and correct line (instead of the ui warnings)
image
 
 Merge when you are OK

PR #42, opened 2023-07-11
remove compilation warnings

atsju:
I removed 54 `[-Wunused-parameter]` warnings in windows build.

I will do one PR per warning type so that it's easier to review. 
538 warnings to go.
I also take note of what I see in the code to fix it later. 
The most important thing I have seen so far=> there are no comments :cold_sweat:. It's normal as Dale has developped and maintained this alone for a long time. If new features are added, they should probably be documented. Adding comments on existing is lower priority.


gr5:
lol.  Yes, very little comments.  Very very slowly I am adding comments to variable declarations to explain things like what unit it is in, just overall architectural things like "this value is in mm here but is also copied to the wavefront object when the object is created and stored in variable m_xxx but it is in pixels"

Most of the actual code I feel doesn't need comments but function explanations and architectural decision comments are very helpful to me.

Another example, "surfacemanager is a single instance class and you can always grab it from the static function "get_instance()".  It contains a list of all the wavefronts created.  It creates the wavefronts from the igrams.  It has many processing functions such as finding best fit zernike's, averaging wavefronts, rotating wavefronts and much more."

A simple comment like that really helps people understand the overall architecture of what can be found where.

dfttools.cpp
```diff
@@ -73,10 +73,10 @@ void DFTTools::setDFTSize(int val){
...
 }
 
-void DFTTools::on_dftSizeSB_valueChanged(int arg1)
-{    QSettings set;
-     set.setValue("DFTSize", ui->dftSizeSB->value());
-
+void DFTTools::on_dftSizeSB_valueChanged(int value)
+{    
+    QSettings set;
+    set.setValue("DFTSize", value);
```
gr5, line 79: So I tested this code the old way and your improved way and both work fine.

atsju: Yes. Basically value was already passed by this signaling function. So there was no need to access it through ui objects.
I did not test the effect of each and every modification I did but this one I did. It's really good that you test the sketchy ones yourself too. Thank you for that.

gr5:
I also built this "the way Dale builds it" and it builds fine.  

Also I noticed that some of the functions were probably auto-created by qt creator so I'm going to do some experiments to make sure the "variable commenting" doesn't confuse the gui editor.  I'm not to worried.

So that's why it's taking so long for me to approve this one.  Plus I have other things on my list for today.  I hope to finish testing today.

gr5 (commented):
Good news.  QT Creator still finds this line of code even though Julien edited it by commenting out the parameter.  I right click on the control "moving" and choose "go to slot.." then choose "clicked(bool)" and qt creator jumps to the line of code properly.
2023-07-14 10_05_58-foucaultview ui @ DFTFringe  JST_fixWarn  - Qt Creator

2023-07-14 10_06_19-foucaultview ui @ DFTFringe  JST_fixWarn  - Qt Creator

2023-07-14 10_06_52-foucaultview cpp @ DFTFringe  JST_fixWarn  - Qt Creator

pixelstats.cpp
gr5, line 1: This seems like a lot of changes - can you explain?  This is the screen that pops up when you click "histogram" above the color 2d graph on the right.  It seems to be working okay but why all these changes?

atsju: I'm on phone so I did not check 100% but these functions are not callbacks/slots/signals. So I removed them because they were not usecalled in code or had no effect. 
I can tell you more tomorrow if you want a more detailed answer.

gr5:
I meant for my comment above to be specifically about pixelstats.cpp.  I'm not sure if my comment was associated with that file or not but that was the intention.

I reviewed everything else.  I'll look at pixelstats.cpp again but your changes look unexpected to me.

gr5:
Okay.  I realize now that the functions you deleted were called from commented code.  Well either you should put those back in and comment out the unused functions, or delete *all* the commented code because if we are saving that commented code then it needs to be in its original form.  We shouldn't randomly delete lines of commented out code from a larger block of commented out code.

atsju:
> Okay. I realize now that the functions you deleted were called from commented code. Well either you should put those back in and comment out the unused functions, or delete _all_ the commented code because if we are saving that commented code then it needs to be in its original form. We shouldn't randomly delete lines of commented out code from a larger block of commented out code.

Yes your comment did link to pixelstat.cpp.  
If it's ok for you I will delete the commented code. Commented code not really had it's place in source. If it's usefull it need to sit in a branch to use for later. Or needs to be well comented for development purpose. Here I have no way to know if it has a purpose or if it has been forgotten. Will do that tomorrow

githubdoe:
Upon review of the current pixelstats code.  Those commented sections were used to debug the code by showing various other plots.  Since it was so complicated to create those debug plots I left them in there to give me an idea how to do that again if ever needed.  Saving them into some other file would be useless as then they would get lost.  I find no harm in them staying where they are.

atsju:
> Upon review of the current pixelstats code. Those commented sections were used to debug the code by showing various other plots. Since it was so complicated to create those debug plots I left them in there to give me an idea how to do that again if ever needed. Saving them into some other file would be useless as then they would get lost. I find no harm in them staying where they are.

Ok. Then I will test this debug code and do modifications accordingly so that it makes some sense

githubdoe:
I will state my comment philosophy here.  I once worked with a respected software engineer who stated the code is the comment.  If you have comments they may not match the code.  Don't use the comment to understand the code.  In general comments are good at high level for complicated concepts which are not obvious.  They can help with units I guess.  DFTFringe expects all linear working units to be in metric units with a few exceptions.  

Comments can be useful for automated document generation.  But I never planned to use that.

githubdoe:
I would suggest to not mess with it.  Leave it alone.  It is unneeded work.  Nothing to be gained.  Just waiting your time and others that need to review the work.

gr5:
One way to leave the code "alone" but still avoid the compiler errors that the function(s) aren't used is to just comment out those unused functions and leave them as is otherwise.

gr5:
On commenting: good commenting is complicated.  The code is likely to change and people will forget to update the comments.  But something like what units a variable is in is one example of many where comments are helpful.  Actually I prefer to spend most of one's descriptive energy into good variable and function names.

But high level comments are another example, as you say, where changing the code shouldn't require one to rewrite the comments.

"code is self commenting" is somewhat true and in many cases I agree.  But overall architectural stuff is helpful.

Interferometry:
How about enabling the debug code in a debug build, disabling it in a release build using compiler directives?

gr5:
I'd just copy Dale's explanation word for word into a comment.  Put it at the start of the commented out code.  And also comment out that function that is called (from the commented code).

pixelstats.h
```diff
@@ -60,9 +60,7 @@ class CanvasPicker: public QObject
     void move( const QPoint & );
     void release();
 
-    void showCursor( bool enable );
-    void shiftPointCursor( bool up );
-    void shiftCurveCursor( bool up );
+    void shiftCurveCursor();
```
atsju, line 63: @gr5 @githubdoe 
- `showCursor` had no effect. That is why I removed it. Please confirm
- `shiftPointCursor` unused and not even defined in cpp
- `shiftCurveCursor` called only in one function with `up=true`, up is only used i commented debug code. That's why I removed the parameter. I can change the debug code to use an internal variable `const bool up`. I don't think this will harm debug code. Please confirm if you are OK with that and I will push the modification.

Please confirm if you are OK with that proposal or if I missed something. I think I changed less thank you imagine (but yes I slightly broke the debug code and must fix this)

atsju:
My opinion regarding comments:
I prepared a quite long answer but I think this would just start a longer debate/argument while I think we already all have the same philosophie. 

Let's just continue to improve all together and discuss on specific code modified in each PR. This will improve code quality each time we touch something.

pixelstats.h
gr5: Oh!  Yes I totally agree.  I don't know why that wasn't obvious yesterday.   These edits look good now.  No "commented out debugging code" will be lost.

gr5 (approved):

PR #44, opened 2023-07-15
fix deprection warnings

atsju:
removed `QString().sprintf()` in favor of [.arg()](https://doc.qt.io/qt-5/qstring.html#arg)

This fixes about 204 occurrences of `[-Wdeprecated-declarations]`.

Also @githubdoe, what is the reason behind using  `sprintf("blabla %s", var.toStdString().c_str())` ? i feel like this is some unnecessary back and forth but I might miss some corner cases

### DIFF
--- a/pixelstats.cpp
+++ b/pixelstats.cpp
@@ -197,7 +197,7 @@ void CanvasPicker::move( const QPoint &pos )
         d_selectedMarker->setLabel( QString("%1").arg(x, 0, 'f') ) ;
     }
     g_centerMarker->setXValue( (g_ub + g_lb)/2.);
-    g_centerMarker->setLabel(QString("PV %0").arg(g_ub -g_lb, 0, 'f'));
+    g_centerMarker->setLabel(QString("PV %1").arg(g_ub -g_lb, 0, 'f'));
 
     /*
        Enable QwtPlotCanvas::ImmediatePaint, so that the canvas has been
--- a/pixelstats.cpp
+++ b/pixelstats.cpp
@@ -189,15 +189,15 @@ void CanvasPicker::move( const QPoint &pos )
     d_selectedMarker->setXValue(x);
     if (idx ==0){
         g_lb = d_selectedMarker->xValue();
-        d_selectedMarker->setLabel( QString().sprintf("%lf", x) ) ;
+        d_selectedMarker->setLabel( QString("%1").arg(x, 0, 'f') ) ;
 
     }
     else {
         g_ub = d_selectedMarker->xValue();
-        d_selectedMarker->setLabel( QString().sprintf("%lf", x) ) ;
+        d_selectedMarker->setLabel( QString("%1").arg(x, 0, 'f') ) ;
     }
     g_centerMarker->setXValue( (g_ub + g_lb)/2.);
-    g_centerMarker->setLabel(QString().sprintf("PV %lf", (g_ub -g_lb)));
+    g_centerMarker->setLabel(QString("PV %0").arg(g_ub -g_lb, 0, 'f'));
 
     /*
        Enable QwtPlotCanvas::ImmediatePaint, so that the canvas has been
@@ -243,6 +243,7 @@ void CanvasPicker::shiftCurveCursor()
         if ( it == curveList.end() ) // not found
             it = curveList.begin();
 
+        const bool up = true;
         if ( up )
         {
             ++it;
@@ -540,7 +541,7 @@ void  pixelStats::updateHisto(){
     muY->setLinePen( Qt::red, 1 );
 
     double ub = g_ub;
-    muY->setLabel( QString().sprintf("%lf", ub) ) ;
+    muY->setLabel( QString("%1").arg(ub, 0 , 'f') ) ;
     muY->setXValue(ub );
     muY->attach( ui->histo);
     histPlot->setSamples(histData);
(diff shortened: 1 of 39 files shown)


githubdoe:
Several std:: routines need c strings as their arguments.  So that sequence is needed for them.  Perhaps there are now other ways to convert QStrings to c strings.

I hate the .arg method that QString now has to format text.  I can never remember it's format.  Since the old method has worked I almost never tried to use the new method.  I was never going to go through the code and change all those print statements.  I'm surprised someone had the patience to do it.  In the future however if I need to print something I'm probably going to use the old method since I don't have time to figure out the stupid .arg syntax each time.

githubdoe:
I was feeling grumpy earlier.  Thanks Julian for all that work.

atsju:
> I was feeling grumpy earlier. Thanks Julian for all that work.

I'm my pleasure, not hard feeling.

sprintf in not exactly my priority but I can't filter out only this exact deprecation. So I had not other choice than to change it to avoid drowning important warning in the list. I'm aiming at 0 warning.

gr5:
For some reason I didn't see this PR until now.  Please merge in latest master to get the LF endings.

Julien this is great work.  I'm going to look at "regions" tomorrow (hopefully!) and if I can check in some fixes for my own PR I plan to do that before reviewing this one.  I want to not just review your changes but also test some of them.

githubdoe:
You probably already know the region code will be found in the igramarea.cpp file.  Lost of code there.  But it must be somewhere related to loadoutline code.

atsju:
This one is not ready for review. I still have some small changes to do. I will rebase of course

gr5:
Is this ready for review?  It says you marked it as ready for review 2 days ago.

atsju:
> Is this ready for review? It says you marked it as ready for review 2 days ago.

Yes it is. 
I try to mark my PR as draft when they are not ready. This is why you though you missed this one.

It's quite a lot of changes I don't think you can test **everything** in a reasonable time.

gr5:
I plan to look at all the changes.  I downloaded and have this version of DFTF running (so handy to have the actions build DFTF for me!!!!) so I can test some of these.  I won't *test* every change but I'll test a few and I'll examine the code for the others.

pixelstats.cpp
```diff
@@ -189,15 +189,15 @@ void CanvasPicker::move( const QPoint &pos )
...
 
     }
     else {
         g_ub = d_selectedMarker->xValue();
-        d_selectedMarker->setLabel( QString().sprintf("%lf", x) ) ;
+        d_selectedMarker->setLabel( QString("%1").arg(x, 0, 'f') ) ;
     }
     g_centerMarker->setXValue( (g_ub + g_lb)/2.);
-    g_centerMarker->setLabel(QString().sprintf("PV %lf", (g_ub -g_lb)));
+    g_centerMarker->setLabel(QString("PV %0").arg(g_ub -g_lb, 0, 'f'));
```
gr5, line 200: Should be %1 but it seems to work fine!?!  I mean it definitely works but should probably still be %1, right?

atsju:
```suggestion
    g_centerMarker->setLabel(QString("PV %1").arg(g_ub -g_lb, 0, 'f'));
```
good catch

pixelstats.cpp
```diff
@@ -243,6 +243,7 @@ void CanvasPicker::shiftCurveCursor()
         if ( it == curveList.end() ) // not found
             it = curveList.begin();
 
+        const bool up = true;
```
gr5, line 246: Why messing with this commented out code?

atsju: It's linked to the discussion we had in a previous PR. 
Function had `up` parameter that has been removed. This "fixes" the commented code in case Dale uses it. It should have been done in the other PR bu I missed it.

gr5: Sound good.

profileplot.cpp
```diff
@@ -429,7 +429,7 @@ void ProfilePlot::populate()
 
     if (m_wf->m_outside.m_radius > 0 && settings.value("GBlur", false).toBool()){
         double val = .01 * (m_wf->diameter) * smoothing;
-        QString t = QString().sprintf("Surface Smoothing diameter %6.2lf%% of surface diameter %6.1lf mm", smoothing , val );
+        QString t = QString("Surface Smoothing diameter %1%% of surface diameter %2 mm").arg(smoothing, 6, 'f', 2).arg( val, 6, 'f', 1 );
```
gr5, line 432: I guess you don't have to escape a double %?  I don't fully understand this but here is what the program actually does:
b

simulationsview.cpp
```diff
@@ -664,13 +664,13 @@ void SimulationsView::on_MakePB_clicked()
         stalkWidth = m_wf->m_outside.m_radius * .1;
         theObstruction = make_obstructionMask(m_wf->workMask);
         m_wf->workMask = theObstruction.clone();
-        showData(QString().sprintf("%f",m_wf->diameter).toStdString().c_str(), theObstruction.clone(), false);
```
gr5, line 667: This is very strange, second parameter isn't used.  But whatever, you kept true to the original intent.  😄

gr5: Oh!  Nevermind.  These must be parameters to showData().  I think I need a break 😄

profileplot.cpp
atsju: excellent. Thanks for review !

gr5 (approved):
Wow that must have been torture to do all those.  It took probably an hour to check all those.  I don't normally check every line of code because usually it's new code but I also didn't want any newly introduced bugs here.  Good work.

atsju:
It took some hours but it's not difficult work once you know the syntax. 

I already reviewed myself and still missed some the one you pointed out. Thank you very much for review

Issue #41, opened 2023-07-08
Linting


Interferometry:
## Is your feature request related to a problem? Please describe.

There has been lots of great work on CI side recently.  Now that we have available builds for three platforms, it might be good to include some basic 'linting' on commit.

I know that when I cloned the repo and saved a file (with no edits), git would show changes due to the Mac vs Windows line endings, for example.

This could initially be used to solve problems with:
- various line endings
- tabs vs spaces
- includes

## Describe the solution you'd like
I'm thinking really basic to begin with. Line endings, duplicate includes, etc...

We can gradually introduce other rules over time, including being able to enforce coding style.

## Describe alternatives you've considered
Er... 
Don't do anything?
Or maybe there is some repo or editor setting that could achieve this?

## Additional context
I've have no idea how to do this, but I know it can be done in a fairly off-the-shelf manner from other projects I've been on. @atsju is this something you have experience of?

atsju:
I have never done it on server side. I have used hooks with uncrustify on client side. 

While I fully agree with you, I think we need to take this step by step and it's not a high priority now. I would push on doxygen format comments more than line ending and spaces. Feel free to propose something if it needs no specific work for historic developers. 

Note that we build only on Linux and windows. Mac is in progress.

Interferometry:
Yes, as I mentioned, it is something to introduce gradually. I know some of it, at least, can be done server side.

This is primarily of benefit to developers, not the CI.

At the most basic level, we want to avoid a sitution where every line in every altered code file gets highlighted as changed. It makes it harder to see what has actually changed.

Interferometry:
Okay so the line endings can be fixed with a repo .gitattributes file. Here is the example GitHub provide.

I'll include an appropriate one in my PR for the macOs version. Not sure why this isn't already set on my repo!

```
# Set the default behavior, in case people don't have core.autocrlf set.
* text=auto

# Explicitly declare text files you want to always be normalized and converted
# to native line endings on checkout.
*.c text
*.h text

# Declare files that will always have CRLF line endings on checkout.
*.sln text eol=crlf

# Denote all files that are truly binary and should not be modified.
*.png binary
*.jpg binary
```

atsju:
Ok your example will at least fix CRLF problems we already dealt with in the past. This would definitely help us.   
It won't remove white spaces on end of line for example or fix indentation. 

Feel free to push a PR and I will try it out to confirm it works as expected.

atsju:
[...]

PR #46, opened 2023-07-16
add .gitattribute to normalize line endings

atsju:
This was a bit helpfull to understand https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/ but there are many other links out there.

Especially: you need to renormalize if you change the content of .gitattributes https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/#a-simple-gitattributes-config

I'm not sure I configured it correctly but `auto` seems to be the consensus. I think files have originally be pushed with CRLF by Dale because of his local git config (normally with auto you checkout CRLF or LF and always push LF so this should not happen) and then someone working on linux build modified the openCV line @gr5 quotes in #41. It has been changed to LF (because pushed from a linux) and this is why this one conflicts more. Conflict also depends on local config of each developper. 
Now all files should be LF in repo. And what happens locally depends on your git config but repo will only have LF.
Note that I tested locally by running `git status` on windows and then in same repo in WSL (linux in windows). Now it works fine. When I did this before while working on the linux build, all files were flagged because line endings has changed. So I suppose it is working as expected. Maybe it needs some more tests.

Let's try and see. @Interferometry what do you think ?


atsju:
@gr5 can we sync on the merge for this PR ?
all files will be flagged as changed. Comparing files from this point to a previous will always require to ignore line endings

so there a 2 reasons to be careful:
- it this one is merged first I need to rebase the 2 other PR to ensure we not introduce bad line endings again. I may be automatic, I don't know.
- maybe it's the right time to execute uncrustify. history will even more be a mess but files will get a cleanup from this point (no automated maintenance though). I will open a draft PR to show you what I mean about uncrustify.

atsju:
look this is what uncrustify can do on just 1 file : https://github.com/atsju/DFTFringe/pull/13/files
of course I can configure almost everything to your taste.

gr5:
No uncrustify please, lol.  Unless Dale thinks it's a great idea, then "no, not at this time".

When you say it will "mess up the history", it will really only mess up a single commit, right?  I don't see that as messing up the history.  And @atsju, you mention two other pull requests? But this seems to be the only active pull request. 

I am 100% for this .gitattributes thing.

@githubdoe - to clarify - this pull request changes EVERY text file (195 files!) in the repository from CRLF line endings to LF line endings but the files  on your hard drive will stay in the CRLF state and for all windows users.  Again - repository will have LF endings and windows users will see CRLF.  This is the github standard when you have people with mac, pc, linux editing the same files.

However, you will need to do a git pull before doing any more editing so I'll ask now: are you in the middle of any edits of DFTFringe at this time?  If so we can wait on this PR until you've pushed all your latest changes.

I personally am in the middle of edits but I don't mind - I can merge this version into my work-in-progress branch once this PR has been accepted.

atsju:
You did an excellent summary of the situation. It does not mess up history, just one commit. It will just be a bit more difficult to compare changes from older to newer files (most tools can ignore the line ending changes, so it's not a big issue). 
Note I never used it before. I hope this Is what will happen but I might be wrong... Let's just see. 

I'm like you, I don't care about modifying my PRs, ready or not. Just don't merge this one and other at same time. Maybe this one first and then I rebase.

About uncrustify you get final word. Of course this is not something you want in middle of something.

Interferometry:
You don't need to recommit all 195 files. Just the .gitattributes file. Once it is in the main repo, the issue goes away.

Next time anyone fetches from the repo, they will also get the new gitattributes and so from then on everything gets handled gracefully.

If you just try to commit an unchanged file (except for the line endings) it even gives you a nice comment explaining why that file is in your commit, and that the file is effectively unchanged.

.gitattributes
Interferometry, line 1: This .gitattributes file looks good. However, there is no need to commit every other source file － in fact you probably don't want to becuause it will unnessarily polute the commit history with code changes that aren't real.

Interferometry (changes requested):
Only .gitattributes is needed. I think its best to leave any unchanged files to be automatically handled by git, and avoid poluting all the other files' change history.

gr5:
So I'm pretty sure all the files have CRLF endings in the repository and they should be LF endings.  So why not just do a massive fix of all of them in a single commit?

githubdoe:
What?  I'm lost.  Why do they need to be only LF.  They are made and used on a windows system.

githubdoe:
So I agree that the .gitattributes seems to be the solution.  I think that was it's intended purpose.  I think then the files are converted on gets and pushes as needed for the system doing the work.  I assume that happens on existing files as well.

atsju:
@Interferometry 
I fully agree with @gr5 ont this :
> So I'm pretty sure all the files have CRLF endings in the repository and they should be LF endings. So why not just do a massive fix of all of them in a single commit?

> If you just try to commit an unchanged file (except for the line endings) it even gives you a nice comment explaining why that file is in your commit, and that the file is effectively unchanged.

So I did some tests and if my understanding is correct it would still commit changes bit by bit. Yes what you propose is more or less transparent for user but not 100% clean
`git ls-files --eol` shows line endings in Index (repo) and Working tree (locally)
image
We have a mix of a bit of everything. For example zernikeprocess.cpp is marked a mixed in index. This means it will be pushed as is with mixed line endings. So the line endings that will be pushed for this file will depend on file and on user and we don't solve 100% the issue. After running `git add --renormalize .` all files not marked as LF in index (as it should be) are pushed with new EOL. Then you never again have line ending changes due to user settings.
image

@githubdoe
> I think then the files are converted on gets and pushes as needed for the system doing the work

Yes files will be LF in repository and when you checkout, they will get the line endings needed for you system (which might be different or not from repo).

Interferometry:
> So I'm pretty sure all the files have CRLF endings in the repository and they should be LF endings. So why not just do a massive fix of all of them in a single commit?

It shouldn't be needed once the .gitattributes file is commited to the repository; it will come down to everyones local repository and inform git how to handle check-outs and commits. I don't think it really matters, I was just thinking about the history.


> So I agree that the .gitattributes seems to be the solution. I think that was it's intended purpose. I think then the files are converted on gets and pushes as needed for the system doing the work. I assume that happens on existing files as well.

Thats my understanding too.

atsju:
@Interferometry  please see above comment, I think we wrote at same time.
also, here is the result of my tests https://github.com/atsju/DFTFringe/pull/14/files

Interferometry:
A single commit would make it easier to verify the change is working okay, and revert if it isn't. So I suppose the issue with the change history is a price worth paying?

gr5:
First of all LF in the repository is the recommended standard for git and also github.  The features in git let you deal with this nicely but the corresponding tools to make it CRLF in the repository don't really exist.  But at the same time having CRLF in the repository is fine as long as every user only uses windows.  Which is not the case.  In addition the windows users have to have never changed that "core.autocrlf" to anything other than default.  Unfortunately I have that set to "true" for the last 15 years so my code possibly has LF in the repository somewhere.

I'm a bit confused about the rest.

Because the files are 99.99% CRLF and only a few lines of code (mostly include statements) have the LF (and possibly also the code I wrote from scratch), then my understanding (I could definitely be wrong!) is that if we *only* change .gitattributes, then if someone changes say just one line of code in X.cpp and they check it in, then every line will be switched to LF and the entire file will show up in diff for that commit.  That would be bad.  Is that true?  Or will the file stay as CRLF in the repository as it is now?

Whereas if we change .gitattributes and update EVERY file in the same commit to be LF then every text file will show up in the commit but the comment on the commit can say something like "only whitespace is changed in this commit as CRLF was changed to LF".

That way we have a single commit showing all text files changed and going forward, no one will notice any issues.  The windows people will automatically see CRLF in their checked out files.  The linux/mac people will see LF and the repository will store LF.

atsju:
> Because the files are 99.99% CRLF and only a few lines of code (mostly include statements) have the LF (and possibly also the code I wrote from scratch), then my understanding (I could definitely be wrong!) is that if we only change .gitattributes, then if someone changes say just one line of code in X.cpp and they check it in, then every line will be switched to LF and the entire file will show up in diff for that commit. That would be bad. Is that true? Or will the file stay as CRLF in the repository as it is now?

from the tests I did (see link above), no this is not 100% true. I did change only one line in files and it did not push whole file. In fact it will likely also depend on how your local tools manage/modify the CRLF themselves. So answer is probably : maybe, maybe not. I'm unsure.

> Whereas if we change .gitattributes and update EVERY file in the same commit to be LF then every text file will show up in the commit but the comment on the commit can say something like "only whitespace is changed in this commit as CRLF was changed to LF".

> That way we have a single commit showing all text files changed and going forward, no one will notice any issues. The windows people will automatically see CRLF in their checked out files. The linux/mac people will see LF and the repository will store LF.

This is correct. Doing this should solve the problem once for all (from my understanding).

gr5:
So @githubdoe , I'll ask again, is this a good time to do this?  Are you part way through some edit?  If not let's do this commit and you can do a pull before you do any more editing.

githubdoe:
I would like someone to make a new version of DFTFringe with the fixes in for the edge mask and any fixes ready for the region issues.  When can that happen?  I would like that to happen soon.  Also my current version has the region bug where if the region is outside the size of the current image it will crash.  I hope George has that fixed.  My idea is if that happens then delete the region.

Sorry if this is a mix of issues but I'm not sure where to comment for something like this.

gr5:
We can release with the current version if you want.  Just tag it (I can probably do that) with a version that is vA.B.C format (e.g. v6.2.1) and the windows installer will automatically appear roughly 15 minutes later in the action menu.

I FINALLY have time starting tomorrow - no more fires to fight - I did look at the region code but did not look for the specific bug of being outside the cropped image.  I can look at that tomorrow.  I *have* been looking at code related to the central obstruction and fixed several bugs but did not release.  I can ignore that branch for now and look at regions instead tomorrow.  Do you want to wait for that?  I see no reason to wait.  The current version has some "edge mask" things fixed (fixes by dale).  However I'm not sure this one edge thing is worth making a big announcement or anything.  So again, just do a pull, tag the branch with "vA.B.C" and do a push and look in the Actions tab on github.  Oh one more thing - you have to somehow copy the installer into github "releases" and say what the latest bug fixes are.  I'm not sure how to do that.

Dale can you also confirm that you aren't in the middle of editing anything so I can pull this PR related to line endings?

githubdoe:
I'm not editing anything.  I think the regions issue is the most important.  It causes the most questions.  So wait till it is fixed to release.

To create a github release you start a new release from some button (probably the releases button.)  You name it and give it a release version number. You drop release notes into it and you drag the installer.exe onto the files box.

gr5 (approved):

gr5:
Okay I confirmed the merge.  Heaven help us.  :)

I don't know if I can do releases but if I fix the region bug then I'll give it a shot.

Interferometry:
Well, the build still works. Looks good!

PR #47, opened 2023-07-19
Fixes some issues with regions that were causing DFTF to crash

### LINKED ISSUE #36: Regions come back after deletion.
Worked on a project that had regions.

Switched to a different project and started to outline igram.  Regions from other project were there.  So I deleted them.  Then I auto traced the outline.  Regions came back.

So two issues. 

 Need some way to decide when to drop the saved regions that are saved in the QSettings.  Lets say on loading a different config or a different base path for igrams.

Figure out how deleted regions don't stay deleted.

gr5:
Fixes #36 

regions were scaled wrong when saved to registry (config memory - different from OLN files) such that they would typically go way off the left edge of the igram and not be visible to users and cause crashes.

Also fixed a bug where regions would be changed to their encompassing square.  But only in debug mode so didn't affect users.

Also changed it so regions are saved to registry (config) when you hit "done" and not when you hit "crop".  This way if the user makes a mistake with a region or other outline they can just reload the igram.

### DIFF
--- a/dftarea.cpp
+++ b/dftarea.cpp
@@ -311,6 +311,17 @@ cv::Mat DFTArea::grayComplexMatfromImage(QImage &img){
         for (unsigned int i = 0; i < igramArea->m_polygons[n].size(); ++i){
             int x = round((igramArea->m_polygons[n][i].x - left) * scaleFactor);
             int y = round((igramArea->m_polygons[n][i].y - top) * scaleFactor);
+
+            // make sure x and y values of regions are inside our matrix
+            if (x < 0)
+                x=0;
+            if (x >= roi.cols)
+                x=roi.cols-1;
+            if (y < 0)
+                y=0;
+            if (y >= roi.rows)
+                y = roi.rows-1;
+
             m_poly.back().push_back(cv::Point(x,y));
         }
     }
--- a/zernikeprocess.cpp
+++ b/zernikeprocess.cpp
@@ -848,7 +848,7 @@ void zernikeProcess::fillVoid(wavefront &wf){
                     if (x < 0 || y < 0 || x >= wf.data.cols || y >= wf.data.rows){
                         continue;
                     }
-                    if (wf.mask.at<bool>(y,x) == 0.){
+                    if (wf.mask.at<uchar>(y,x) == 0.){
                         ux = (double)(x - midx)/rad;
                         uy = (double)(y - midy)/rad;
                         rho = sqrt(ux * ux + uy * uy);
--- a/zernikeprocess.cpp
+++ b/zernikeprocess.cpp
@@ -848,7 +848,7 @@ void zernikeProcess::fillVoid(wavefront &wf){
                     if (x < 0 || y < 0 || x >= wf.data.cols || y >= wf.data.rows){
                         continue;
                     }
-                    if (wf.mask.at<uchar>(y,x) == 0.){
+                    if (wf.mask.at<uint8_t>(y,x) == 0){
                         ux = (double)(x - midx)/rad;
                         uy = (double)(y - midy)/rad;
                         rho = sqrt(ux * ux + uy * uy);
--- a/igramarea.cpp
+++ b/igramarea.cpp
@@ -1962,8 +1962,8 @@ void IgramArea::saveRegions(){
     for (int i = 0; i < m_polygons.size(); ++i){
         regions << "Poly";
         for(int j = 0; j < m_polygons[i].size(); ++j){
-            regions << " " << QString().number(m_polygons[i][j].x) << ","<<
-                       QString().number(m_polygons[i][j].y);
+            regions << " " << QString().number(m_polygons[i][j].x+cropTotalDx) << ","<<
+                       QString().number(m_polygons[i][j].y+cropTotalDy);
         }
         regions << "\n";
     }
@@ -2030,6 +2030,7 @@ void IgramArea::crop() {
     x = igramGray.width()/2;
     y = igramGray.height()/2;
 
+
     for (int i = 0; i < m_polygons.size(); ++i){
         for(int j = 0; j < m_polygons[i].size(); ++j){
             m_polygons[i][j].x -=crop_dx;
@@ -2037,7 +2038,6 @@ void IgramArea::crop() {
         }
 
     }
-    saveRegions();
 
     m_outside.translate(QPointF(-crop_dx,-crop_dy));
     cx = m_outside.m_center.x() + crop_dx;
@@ -2262,6 +2262,7 @@ void IgramArea::writeOutlines(QString fileName){
             file << std::endl;
         }
     }
+    saveRegions(); // save regions to registry also
     if (m_edgeMaskWidth != 0){
         mirrorDlg &md = *mirrorDlg::get_Instance();
         file << "\nEdge Mask width" << std::endl << md.aperatureReduction << std::endl;


dftarea.cpp
```diff
@@ -311,6 +311,17 @@ cv::Mat DFTArea::grayComplexMatfromImage(QImage &img){
...
+
+            // make sure x and y values of regions are inside our matrix
+            if (x < 0)
+                x=0;
+            if (x >= roi.cols)
+                x=roi.cols-1;
+            if (y < 0)
+                y=0;
+            if (y >= roi.rows)
+                y = roi.rows-1;
```
atsju, line 323:
```suggestion
            if (x < 0)
                x=0;
            else if (x >= roi.cols)
                x=roi.cols-1;
                
            if (y < 0)
                y=0;
            else if (y >= roi.rows)
                y = roi.rows-1;
```

dftarea.cpp
```diff
@@ -311,6 +311,17 @@ cv::Mat DFTArea::grayComplexMatfromImage(QImage &img){
         for (unsigned int i = 0; i < igramArea->m_polygons[n].size(); ++i){
             int x = round((igramArea->m_polygons[n][i].x - left) * scaleFactor);
             int y = round((igramArea->m_polygons[n][i].y - top) * scaleFactor);
+
+            // make sure x and y values of regions are inside our matrix
+            if (x < 0)
+                x=0;
```
atsju, line 317: This is only a suggestion:
```suggestion
            if (x < 0)
            {
                x=0;
            }
```
Matter of taste but not only. I believe this is MISRA rule 15.6 meaning enough people made mistakes with such code that defensive style is recommended for everybody.

zernikeprocess.cpp
```diff
@@ -848,7 +848,7 @@ void zernikeProcess::fillVoid(wavefront &wf){
                     if (x < 0 || y < 0 || x >= wf.data.cols || y >= wf.data.rows){
                         continue;
                     }
-                    if (wf.mask.at<bool>(y,x) == 0.){
+                    if (wf.mask.at<uchar>(y,x) == 0.){
```
atsju, line 851: are you sure about this line ?
Because :
https://github.com/githubdoe/DFTFringe/blob/1daf63a5dba50d9512b41a11263c122a84df85af/wavefront.h#L31

I'm doing C++ only for 1 year and not full time so my understanding might be wrong but shouldn't it be :
```suggestion
                    if (wf.mask.at(y,x) == false){
```

or maybe 
```suggestion
                    if (wf.mask.at<CV_U8>(y,x) == 0.){
```

For me the mask and workMask are managed in a inconsistent way unless I miss something about C++ or `cv::Mat_`. They are matrix of bools ? Then why do they store different values than true and false ? 
https://github.com/githubdoe/DFTFringe/blob/1daf63a5dba50d9512b41a11263c122a84df85af/surfacegraph.cpp#L263

atsju (commented):
The "closes #36 must be either in PR main comment/description or in the text of commit. 
PR title and subsequent PR comments are not automagically linking 
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue
I suggest you change title and add the closes in PR main comment

zernikeprocess.cpp
gr5: This line of code was very frustrating and took me a few hours to get working (this was a month or two ago when I was doing a similar edit).  It would work fine in release but when I stepped through with the debugger it wasn't working.  I tried changing the "0." to "false" and that didn't help.  I finally just copied code that Dale had elsewhere (the <uchar> thing) and it worked so I left it.  I did not try leaving out the "type" and I did not try "CV_U8".

Why does C++ use a whole byte for booleans?  I think it's more efficient (speed - not memory size of course).  It's quite ugly to assume that a bool is a whole byte long so I try to never do that but on mac/linux/pc it will always by a byte.  On some future platform, who knows?  I don't think c++ standard specifies that it is exactly one byte.  I'm not sure.

gr5: Regarding the "if" statements.  My actual perferred way to do this is to make it more readable than more bug free.  A strange preference!  I know!  

I have been reading other people's code since I was 10 years old I value readability the most.

My actual preferred way would be to put each command on the same line as the if statement (not always - just in this specific code example) so more like:
if (a<0) a=0;

I've seen people do:
a=max(a,0);
but find that very difficult to read as people get negative values and magnitudes confused.  I just think it's harder to read the min() and max() method.

I know you like braces around a single line if statement as you mentioned it also a day or so ago but I don't. Tabbing is more important.  A good editor or compiler should warn me if the tabs are off.  Right now there are too many warnings but fortunately you may be fixing that.

I did think about using the "else" statements.  That should in theory improve the efficiency of the code but I felt the gain in efficiency was too small to be worth hurting the readability somewhat.

gr5: Believe me, I think about style and readability and possible bugs all the time.  One thing I do is think "if I were the boss, would I tell this programmer to change this?" and if the answer is yes then I change it.  I don't always remember to use that trick but usually.  Always there is this voice "you should add braces", "you should use a better variable name" and so on.  I usually listen to the voice and "do it the right way".

gr5: Back to the boolean - I tried other things as well.  I tried "== false"  "!= true".  Eventually I just copied Dale's other code that was working.  Maybe bool is more than one byte and it was indexing to the wrong location of the matrix.  But then why is bool a different sizeof() value in debug versus release?

atsju: boolean is indeed undefined https://en.cppreference.com/w/cpp/language/types

I did a bit more research after the comment and my best guess would be `wf.mask.at(y,x) == false` without the static cast and with false. However I did not debug through the code, so if you tell me what you did is working I believe you. I would just prefer to understand why it works.

Additional guess: the mat should contain bool but because of poor management elsewhere it contains other values than true or false that mess things up ?

--------

About the braces, I also prefer the one liner you propose in comment. Really this is matter of taste, I won't push for this change.   
About the `else if` however, I find it more logical this way and not only for improving performances. Also matter of taste.

For sure I know I'm very picky on naming and other things, my colleagues are too :) I just need some time to adapt to you as my peer.

atsju: +1 About the bool, while explaining to @hish15 at work, `wf.mask.at<uchar>(y,x) == 0.` this is comparing a uchar to a double 0. ? I'm not sure what this does.  https://godbolt.org/ tells it's the same as compare to int 0 but this little dot might explain the difference between debug and release ?
still unclear what happens

gr5: I'll test this today.  This is interesting.  The period indeed is ugly as it makes it a double.  I almost deleted the period but was just trying to get this working and was frustrated at the time (again - was a few months ago).

But now I want to try your ideas.  You may run into this or related code when you fix all the warnings.  It's possible the other code where the matrix was loaded will have a warning.

gr5: Is the <uchar> a static cast?  I thought it was as template function?  I still think it's a template function but didn't look at the definition of the "at" member function.

Anyway I'll test the code today.

atsju: in vectors/arrays, `at` is same as `[]` but is supposed to make boundary check. What is done in openCV I don't know.

My local expert @Hish15 :) explained me this `<>` in indeed not a cast (I wronly though it was as https://docs.opencv.org/3.4.15/df/dfc/classcv_1_1Mat__.html#a098ec36001b3998a7511e3c9169f93c8) 
But here it is a template specification. So here it should be either bool (with respect of declaration https://github.com/githubdoe/DFTFringe/blob/1daf63a5dba50d9512b41a11263c122a84df85af/wavefront.h#L31) OR nothing. But anything different from bool or nothing is not clean and might even be UB (undefined behavior).

so the only "correct" syntax would be
`wf.mask.at<bool>(y,x) == false`
or
`wf.mask.at(y,x) == false`

gr5:
So the mask was created and filled with uchar and it fills the values with 255 or 0.  That's why it must be uchar.  the <bool> line of code was a mistake Dale made at some point but the compiler didn't care.

I did try .at with no template and it got a compiler error.  I did change the "0." to "0" and that works fine so I'll change that and push.

The reason it's uchar versus bool is (I believe) because it's meant as a mask with variable transparency.  I think opencv supports mask features where you can have variable transparency of a visual mask and Dale is kind of stealing this concept but he is masking out where the outer circle is so the processing only occurs with points in the igram.  The inner obstruction feature, the "ignore regions" feature and the edge mask feature all fill in this mask.

atsju:
> mask was created and filled with uchar

then this probably need to be changed to uchar also ? Or did I miss something ?https://github.com/githubdoe/DFTFringe/blob/1daf63a5dba50d9512b41a11263c122a84df85af/wavefront.h#L31

And -personal taste trigger warning- could we use uint8_t and so on for all integer values ?
Since C++11 these are even officially defined https://en.cppreference.com/w/cpp/types/integer it's not some sort of header that only some compiler will define and might have different implementation.

githubdoe:
My network is back now.  I have been offline for awhile.

The Mat.at<Bool or UChar>(x,y) is how opencv lets you access the matrix individual components.  

There are times when I want to use it as a bool and just do a logical test.  There are other times when I need to use the mask to be inverted in value or to multiply it piece wise against the data (for reasons I now forget).    I no longer know if comparing it to 0. is a typo or a trick.

atsju:
You should not make changes "only" to make me happier ;) Thank you for having investigated this tricky line after my comment. 

There is enough important work to not debate 3h about syntax, it's not what I want either. If you tell me it works and you had a check, I believe you.

gr5:
I don't know why Dale but the at<bool> code usage when defined as a <uchar> doesn't work when compiling in debug mode.  It just doesn't work at all.

Also if you grep for all cases of "workMask.at" I noticed every reference is (y,x) except one which is almost certainly a bug.  It seems wavefront.mask should always be <uchar> and wavefront.workMask should always be <bool> as that is how they were defined.  I don't think you should mix and match.

Also there may be 2 more bugs with mask.at<bool> in zernikeprocess.cpp but I didn't want to mess with that.  Release version works fine and I didn't particularly want to figure out how to test it to know for sure if it matters in debug versus release mode.

gr5:
Dale if you could review the 17 or so lines of code changes I would appreciate it.  Just click on the files tab above to see the changes (along with @atsju  comments mixed in).

atsju:
@gr5  you can close the PR conversations as resolved to ease dales review (as they are). I masked some but I have no right to "resolve" as I'm not maintainer.

gr5:
> then this probably need to be changed to uchar also ? Or did I miss something ?
> 
> https://github.com/githubdoe/DFTFringe/blob/1daf63a5dba50d9512b41a11263c122a84df85af/wavefront.h#L31
> 

Oops.  I missed that one.  Yeah that should be uchar also.  Interesting.  Fixing this is very very simple.  Testing, not so much.  I'd rather not fix all of these right now but I could if Dale wants me to.  Fixing will take 5 minutes.  Testing... not sure and I may miss something.

githubdoe:
Does "No rights as maintainer" mean you need a right added and if so I will but you have to tell me how?

It's 2:30 am and my mind is mush.  I looked a the code to be reviewed but brain would not process it.  You all know how poor of a proof reader I am.  I see what I want to see.  Same goes for code reviews.  I'm use to letting the computer find my mistakes if the compiler can.  Then I try to dream up test cases to see if the code works.  I'm too old school to know how to write automatic test cases for a GUI.  So it is hit and miss when I dream up manual test cases.

I'm stuck in the past C++ without   the version add on like 11 and so on.  So in general you won't find me using any of those new features except for a tiny tiny few because I just don't know them.  All the C++ I learned was from the original C++ book.  I actually read the whole thing.  Go ahead and use any of the new stuff as appropriate however.

atsju:
> It's 2:30 am and my mind is mush.

We are not in a hurry I think this can wait

> Does "No rights as maintainer" mean you need a right added and if so I will but you have to tell me how?

For clarification, @gr5 has more rights than I have on you repo. In fact I have 0 more rights than anybody. However I do not feel the need to get elevated privileges as it doesn't lock me out from contributing. And with great power comes great responsibility.
Today I can NOT merge PRs, create tags and releases, close PR comments, ... If you feel like it would be useful to have a second person with same rights as @gr5 to do these things then feel free to add me but it's not a requirement on my side.

gr5:
If you don't care then I'll just do the merge.  I'm reasonably confident in the code changes.  I did test it all.

If you want to look the code changes over, it's only 17 lines of code.  The idea is that because you wrote the code originally, you may know something I didn't consider.  You don't have to always look at every line and be 100% sure it will work.  The idea is that you might notice something because of your experience.

Once I do the merge I'll try to see if I can create a release.  If I can't I'll let you know.

githubdoe (commented):
I have trouble reviewing code without seeing the whole file to see what routine diffs are in and their context.  I could not find a way to show the whole file as a diff.  There must be a way I'm sure.

githubdoe (commented):
I think I have discovered the expand up and down buttons that shows the whole file.

githubdoe (approved):

PR #52, opened 2023-07-22
fix some compilation warnings [-Wunused

atsju:
### Before PR on my local build:
[-Wunused-local-typedefs] 32
[-Wunused-variable] 40
[-Wunused-but-set-variable] 25
[-W 333

### After PR on my local build:
[-Wunused-local-typedefs] 32 👎 
[-Wunused-variable] 0 👍 
[-Wunused-but-set-variable] 10 👎 
[-W 276 (meaning 2 other random warnings disapeared in the process)

- Remaining [-Wunused-local-typedefs] all come from armadillo. FYI this has not been fixed in latest version 12.6.0 but I opened an issue and next release will work fine https://gitlab.com/conradsnicta/armadillo-code/-/issues/239 will be corrected in #53 as soon as Armadillo publish a release (usually 1-2 month)
- Remaining use_alpha set but not used [-Wunused-but-set-variable] => I believe this is a compiler bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96311 . Would probably be fixed if we upgrade compiler (for this we need to update QT).

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -1078,7 +1078,6 @@ void MainWindow::batchProcess(QStringList fileList){
     int last = fileList.size()-1;
 
     int ndx = 0;
-    VideoWriter *vw;
     int cnt = 0;
     int width, height;
     QString videoFileName;
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -63,20 +63,20 @@ MainWindow::MainWindow(QWidget *parent) :
     ui->setupUi(this);
 
 
-    const QString toolButtonStyle("QToolButton {"
-                                    "border-style: outset;"
-                                    "border-width: 3px;"
-                                    "border-radius:7px;"
-                                    "border-color: darkgray;"
-                                    "font: bold 12px;"
-                                    "min-width: 10em;"
-                                    "padding: 6px;}"
-                                "QToolButton:hover {background-color: lightblue;"
-                                        " }");
-    QWidget *rw = ui->toolBar->widgetForAction(ui->actionRead_WaveFront);
+    //const QString toolButtonStyle("QToolButton {"
+    //                                "border-style: outset;"
+    //                                "border-width: 3px;"
+    //                                "border-radius:7px;"
+    //                                "border-color: darkgray;"
+    //                                "font: bold 12px;"
+    //                                "min-width: 10em;"
+    //                                "padding: 6px;}"
+    //                            "QToolButton:hover {background-color: lightblue;"
+    //                                    " }");
+    //QWidget *rw = ui->toolBar->widgetForAction(ui->actionRead_WaveFront);
     //igramArea = new IgramArea(ui->tabWidget->widget(0));
 
-    rw = ui->toolBar->widgetForAction(ui->actionSubtract_wave_front);
+    //rw = ui->toolBar->widgetForAction(ui->actionSubtract_wave_front);
     //rw->setStyleSheet(toolButtonStyle);
     me = this;
     this->setAttribute(Qt::WA_DeleteOnClose);
@@ -1646,7 +1646,7 @@ void MainWindow::on_actionCreate_Movie_of_wavefronts_triggered()
 
 //                pd.setLabelText(name);
 //                QApplication::processEvents();
-//                wavefront *wf = m_surfaceManager->readWaveFront(name,false);
+//                wavefront *wf = m_surfaceManager->readWaveFront(name);
 
 //                m_surfaceManager->makeMask(wf);
 //                m_surfaceManager->generateSurfacefromWavefront(wf);
--- a/zernikeprocess.cpp
+++ b/zernikeprocess.cpp
@@ -1118,7 +1118,7 @@ void ZernikeSmooth(cv::Mat wf, cv::Mat mask)
 
 
 arma::mat zernikeProcess::rhotheta( int width, double radius, double cx, double cy,
-                                   double insideRad, const wavefront *wf){
+                                   const wavefront *wf){
     bool useMask = false;
     if (wf != 0){
         useMask = true;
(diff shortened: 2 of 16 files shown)


mainwindow.cpp
```diff
@@ -63,20 +63,20 @@ MainWindow::MainWindow(QWidget *parent) :
...
+    //                                "min-width: 10em;"
+    //                                "padding: 6px;}"
+    //                            "QToolButton:hover {background-color: lightblue;"
+    //                                    " }");
+    //QWidget *rw = ui->toolBar->widgetForAction(ui->actionRead_WaveFront);
     //igramArea = new IgramArea(ui->tabWidget->widget(0));
 
-    rw = ui->toolBar->widgetForAction(ui->actionSubtract_wave_front);
+    //rw = ui->toolBar->widgetForAction(ui->actionSubtract_wave_front);
     //rw->setStyleSheet(toolButtonStyle);
```
atsju, line 80: I was unused if I should delete or only comment this...
I'm not used to Qt enough to understand what it does

zernikeprocess.cpp
```diff
@@ -1415,21 +1411,15 @@ std::vector<double>  zernikeProcess::ZernFitWavefront(wavefront &wf){
     'calculate LSF matrix elements
     */
 
-    Settings2 &settings = *Settings2::getInstance();
```
atsju, line 1418: Please double check if this one has a side effect.  I'm unsure with this `new` inside call and static member...

gr5: Dale has many classes that are only a single instance ever.  Only one object ever created of that class.  What he likes to do is give it a static function called "getInstance()" which grabs that single instance of the class.  The very very first time this function is called, the single instance is created.

Deleting this line of code is absolutely fine.

atsju: This was my understanding too but after discussing with my colleague so that they explain it to me.

It should probably not be done this way in 2023 but we did not have time to discuss it further. 
It might probably be a global instance or an instance living in `main()`. `getInstance()` could still exist but there would be no memory leak (I agree there is no **problematic** leak here but there is still a `new` without `free` which is not OK).

Well this is just to share my though with you. This is not a priority fix and the details of how to do it cleanly are not clear for the moment.

gr5: This technique is used in many places and I think it's fine.  There are probably 10 or 20 classes with this function.  Classes that will only ever exist one time.

Then of course there are other classes like wavefront.  There are many of those.

Yes they could all be created in main() and deleted on program exit but not sure that matters.  I guess that would assist in checking for memory leaks as memory leak tools check to see what never gets deleted when the program is complete.

I just built the master branch and that is down to 267 warnings.  Progress!

atsju: > I guess that would assist in checking for memory leaks as memory leak tools check to see what never gets deleted when the program is complete.

exactly.

githubdoe:
Singletons:
There are many classes that should be a singleton that other classes need access to but were created by Mainwindow.  Eventually it became awkward to make each public in Mainwindow and then include Mainwindow.  It created circular includes.  Eventually I started making those singletons with the get instance.

Memory Leaks:
Yes I'm sure there are leaks and back in the days when Visual Studio was more affordable and I used MFC it had memory leak detection in the IDE.  When I switched to Qt that was no longer available and I have never found the method for Qt on windows  without Visual Studio to get leak detection.    I missed it greatly and forged ahead.  

So any help there would be very good.  Also if one leaves DFTFringe running on a windows machine it eats up CPU usage.  Why?

Armadillo was once needed by Mike Pecks PSI and/or Zernike generation software.  I never actually implemented all of what I intended of Mike stuff.  So I have lost track of what of it was used.  I think there may have been some templated objects.  I'm likely never to do the original intention.

mainwindow.cpp
gr5: I think it's fine.

gr5 (commented):
Everything looks good.  All the code changes look fine.  But when I build this on my laptop (I usually build on my pc but I'm not at home now) I can't load igrams.

I don't know what's going on but I'll try building this on my pc tomorrow.

The instance built by "actions" works fine so I just have the wrong dll somewhere or something.

gr5:
Dale thanks for the background information.

It didn't occur to me but of course you can do both: have singletons that have a "new" command in getInstance() and also instantiate all of those singletons in main() (and delete them upon exit).  In fact also checking for null in the getInstance() is potentially just a precaution.  So I have absolutely no issues with that way of doing things.

githubdoe:
Not all singletons need to be instantiated in main but instead only the first time they are needed.  Thus the null check to keep them from being instantiated again.

githubdoe:
In case you did not know there are books about useful software patterns that occur and are particular useful in object oriented languages.  Singleton is one such pattern.  That was very popular when I initially started writing OpenFinge and DFTFinge.  I don't know the state of "Patterns" these days in software engineering.

atsju:
Dale, thanks for the background infos. This really helps. 

About the singleton design pattern I simply want to share 2 infos:

1. Popularity of this pattern has decreased a lot. Now, this pattern should be avoided in most situations. I like this website but all recent ressources are agreeing on this point. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Ri-singleton
2. When using it, it should be updated to modern C++ (>=C++11). The new design pattern is still roughly the same but it is safer.

Standards are evolving, it doesn't mean the code is bad. I think it is important that we know it should be avoided but the changes would probably be quite complex. I would not see this change as high priority except if we start chasing memory leaks with Valgring or other tools. 

I also had an interesting/long discussion with ChatGPT https://chat.openai.com/c/97c005b1-0940-4985-9978-076532dab7ad
But this is more for your information. I'm sure there are plenty of better ressources on the internet. No need to read it all. If you still read it, note that alternative approach 1 is probably the cleanest even if not the easiest.

githubdoe:
DFTFringe's singletons are classes that contain mostly temporary state information that needs to be accessible to many different classes.  This data needs to remain around and stable as long as the app lives and then closing the app deletes it.  Since the deletes happen by the app close no specific delete is needed nor desired.  I would hope the leak detector understands that case.  IIRC the old Visual Studio leak detector did.

Each singleton is a manager of that data and can give clients views into it.  There can only be one manger for each related state data store.

In the old days of C they would have been global variables but are much better grouped by classes now I think.

PR #53, opened 2023-07-22
Update armadillo to v12.6.1 to remove build warnings

atsju:
as discovered in #52 there are 32 warnings due to Armadillo

> Remaining [-Wunused-local-typedefs] all come from armadillo. FYI this has not been fixed in latest version, I openend an issue 
https://gitlab.com/conradsnicta/armadillo-code/-/issues/239

Move from 134 to 102 warnings

TODO later: QWT build should go to a generic folder name to simplify update  (same for  openCV but not possible due to dll name)

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -8,7 +8,7 @@ on:
   workflow_call:
 
 env:
-  armadillo_version: 12.6.0
+  armadillo_version: 12.6.1
   QWT_version: 6.1.6
   openCV_version: 4.6.0
   QT_version: 5.15.2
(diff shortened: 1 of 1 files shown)


atsju:
> Since Armadillo isn't used and only one or two types are used. Maybe we should just get rid of all of armadillo and create the type definitions ourselves? I'll try removing includes to armadillo and see how many types are undefined...

As Armadillo is well maintained, I don't think it's a priority. However you can try to remove it, it could simplify build if we are not required to add too much custom.

Interferometry:
I had a brief look a few days ago, and could find virtually nothing in the compiled code. A bit is pulled in by the 'About' screen - the version numbers I guess. And there is also something related Zernike calculations. But as @gr5 mentioned before, it looks like it could just be some of the definitions in the headers.

It's always nice to reduce the number of dependancies, though it might be better to spend time elsewhere unless it a quick task?

gr5 (approved):

PR #58, opened 2023-07-31
remove most remaining warnings for windows

atsju:


igramarea.cpp
```diff
@@ -453,13 +453,13 @@ cv::Point2d IgramArea::findBestOutsideOutline(cv::Mat gray, int start, int end,i
     MainWindow::me->progBar->setRange(0, cnt);
     cnt = 0;
     int goodCnt = 0;
-    double rmean;
```
gr5, line 456: Rats!  I have been editing mostly just igramarea.cpp and I decided to get rid of all the warnings and so I did basically the exact same edits and will get merge conflicts when I create the PR.  I guess I'll undo my "warning" related edits before I commit. sigh.  lol.

gr5 (approved):

igramarea.cpp
atsju: Oh sorry. You could also have let me fix the merge conflicts. 

Not sure you know it but you can just run `git mergetool` to help you fix the conflict. I have configured it to use meld tool but you can use any other editor you like (prefer one with 3 view for merge).

PR #60, opened 2023-07-31
never do `using namespace std`

atsju:
This is bad practice. Please never ever do `using namespace std;` again :)

I'm not removing for style but because it bytes me for what's coming next.


githubdoe:
IIRC
Left over from legacy C.  Years ago one only had to include "math.h" to get max.  Then when switching to Qt I think they had one.  Then I think they removed theirs.  Some got changed to fully qualified while others were left hanging and worked because there was one in std.  Yes over the years I liked to use the fully qualified names but the using namespace got left.

PR #61, opened 2023-07-31
remove warnings from rule of 3/5/0

atsju:
fix some [-Wdeprecated-copy] warnings only showing on newer compiler
https://en.cppreference.com/w/cpp/language/rule_of_three

For bezier.h, I noticed original repo and opened a PR there too https://github.com/oysteinmyrmo/bezier/pull/10
So our file would be identical to their latest version.

**as for wavefront, I fixed the warnings but not the code.**

copy constructor is not copying following fields and I'm not sure it's normal:
```
QString name;
QVector<std::vector<cv::Point> > regions;
bool regions_have_been_expanded;
```

Also the copy constructor does a deep copy of cv::Mat while the copy operator does not. I'm not sure this is OK or not. I suppose Dale took care of using `new` operator where he needed deep copy but I'm not sure this is correct approach.

Note that for the destructor, I think a default one would work fine. `clear` is supposed to be called automatically in cv::mat destructor.

### DIFF
--- a/bezier.h
+++ b/bezier.h
@@ -159,11 +159,6 @@ namespace Bezier
                 this->normalize();
         }
 
-        Vec2(const Vec2& other)
-            : x(other.x)
-            , y(other.y)
-        {}
-
         Vec2(const Vec2& other, bool normalize)
             : Vec2(other.x, other.y, normalize)
         {}
--- a/wavefront.cpp
+++ b/wavefront.cpp
@@ -33,7 +33,8 @@ wavefront::~wavefront()
     nulledData.release();
 
 }
-wavefront::wavefront( wavefront &wf): data(wf.data.clone()),
+wavefront::wavefront( const wavefront &wf): 
+    data(wf.data.clone()),
     nulledData(wf.nulledData.clone()),
     mask(wf.mask.clone()),
     workData(wf.workData.clone()),
--- a/wavefront.h
+++ b/wavefront.h
@@ -25,7 +25,11 @@ class wavefront
 public:
     wavefront();
     ~wavefront();
-    wavefront( wavefront &wf);
+    wavefront( const wavefront &wf); // copy constructor doing deep copy of cv::Mat
+    wavefront( wavefront && ) = delete;	// move constructor, deleted because unused
+    wavefront& operator=( const wavefront & ) = default; // copy operator not doing deep copy of cv::mat
+    wavefront& operator=(wavefront &&) = delete; // move operator, deleted because unused
+
     cv::Mat_<double> data;
     cv::Mat_<double> nulledData;
     cv::Mat_<uint8_t> mask;


gr5 (approved):

gr5:
not copying the name - that's almost certainly fine.  Everytime dale does a copy he is creating a new wavefront and he always gives it a name.

Not copying the regions - it's probably fine.  Regions are more important earlier in the process - by the time you are creating a wavefront you should also be filling in all the regions with zernike-smoothed data and at that point you can delete all the regions.  That's my opinion anyway.  And if regions are gone then I suppose the bool is no longer needed.

It certainly wouldn't hurt to copy that bool across.  I think the code is fine as is.  I'm more worried about copying matrixes or not.  Usually dale copies the wavefront to a new one (for example if you do a rotate you get a new wavefront - or an average).  At this point (after a rotate or average) you don't want any regions to still exist. And I think he typically creates all new matrixes (data, workdata, mask, workmask, etc).

Any thoughts, Dale?

atsju:
Personally I would always do a deep copy for matrixes and also copy names and everything else. 

- You can still change the name later
- The code always do the same thing, easy to maintain (here copy operator != copy constructor)
- STL (std::) only makes deep copies, so we would align with it
- If we really want a light copy for -reasons- then we could use a function with an explicit name to do it (instead = or constructor )

gr5:
When you say "deep" versus "light" copy I assume "deep" means that member classes are duplicated.

Does light mean your members point to the other object's members (through pointers or references)?  Light would seem dangerous.

gr5:
To give this all context, I once again refer to the "rotate" example where when you rotate a wavefront it creates a new rotated wavefront (and adds some text to the name of the wavefront and adds it to the list of wavefronts in the gui).

You want to create a completely independent wavefront which does not have any links to any data in the unrotated wavefront.  In fact users can delete the unrotated wavefront without the code having to fix any member variables.

Dale is using the "=" syntax to create the new wavefront and copy some of the data over (but again - most of the data is replaced anyway so maybe it doesn't matter - or maybe it does matter - not sure without looking at the code yet again).

atsju:
yes if you look at documentation of cv::Mat (from my understanding) the default copy operator/constructor does not call clone() and thus the matrix is not really copied. Our copy constructor does the clone. But our assignement operator being the default one it does not.

~~Maybe I need to look at it again now that I understand better. I probably need to double check to ensure what I'm telling you now is correct.~~
https://docs.opencv.org/4.6.0/d3/d63/classcv_1_1Mat.html#aed1f81fe7efaacc2bd95149cdfa34302
I had a look again and maybe it's doing copy on write. So maybe I'm wrong, the copy is light but will be deep as soon as it's touched. If this later is true then I would recommend removing the copy constructor as the default one would perform equally well and we should not write ctors that do same as default one.

In any case, here in this PR I did not change anything to the code. Only removed the warning by expliciting the 5 operators. So you can "safely" merge and we will discuss further in #62 some day.

> Dale definitely uses the "=" sign to copy wavefronts. My understanding is not fantastic but in the older code I believe it uses the copy constructor when he does that because he never defined operator=
> 
> But now you explicitly are creating an operator= which does "default". I'm not sure I know what that means. Will it still use the copy constructor like it has been?

I think you are wrong. When copy operator is not explicit (like it was) it does just the same as now (the default one). It does not use the copy constructor instead. And the default assignement operator of cv::mat is not doing the clone. Thus, = will perform a light copy. But copy ctor does the clone so it does a deep copy.

gr5:
Wow!  Okay so I watched half of that first video.  So I never heard of this "copy on write" thing.  It makes c++ more like java.  QString implements this as well - it has a reference counter.  I didn't know!  It also makes writing code using strings so much easier in c++.

So I have 2 comments about this "copy on write" feature.

**QString**

Regarding the QString syntax with the const &.  It is no more efficient to use const & because QString has the "copy on write" and reference counting features built in!  I had no idea.  So that means if the function that receives the string doesn't modify the string then only a reference is passed (actually a pointer I assume - sort of the same thing).  So it's super efficient! no copy made.  But if the function receiving the QString as a parameter *does* make an edit to the string, a copy is made first!!! This is amazing.  I don't think std::string does this?  I don't know.

Then watching more of the video there are many pitfalls using this "copy on write" style of containers.  All new pitfalls to learn.  ugg.  But we are using them already with cv::mat so maybe I just need to learn all the pitfalls so I know to look for them.  Some of the pitfalls are related to inefficiencies.  Some are pure bugs and may explain my problems with <bool> versus <char> when getting confused about the mask array.

Anyway this is new to me.  I feel like the c++ people looked at java and said "how can we make c++ more like java in a good way?"

Anyway the reason people don't like c++ is because it has like 40 years of incremental features added and the old features never go away (and some of them like malloc() probably should!) and recommended ways to use them and the pitfalls keep changing with each new wave of programming style.

**cv::mat**

SECOND POINT: if cv::mat has this same deep copy and reference feature then that's great.  We don't have to do cloning if we don't feel like it and the matrix will auto copy if needed.  That's pretty amazing.

gr5:
Okay so I'm going to approve this pull but no releases until it's tested thoroughly.  Because I am doing code fixes and always using the latest master, I'm sure if we create any bugs I'll find them soon enough.

atsju:
I think you really need to look at https://www.milosolutions.com/en/blog/what-is-copy-on-write/

I'm unsure if passing by copy will trigger the copy or not.

gr5:
read.  Done.

So my take on this is that I should run clazy over the code howerver...

1) I believe for the most part, Dale does not pass wavefronts and matrixes and so on through function calls and instead is more likely to pass a pointer to one.  So I don't think there are huge inefficiencies.
2) QString I don't care about.  They are small.  I don't have to worry about efficiency.  There are millions of operations being performed on qstrings (as there are on matrixes).

Having said all that I think I will be keeping an extra careful eye on when certain variables are copied!  clazy might be helpful but I'm afraid it's going to help us save a few milliseconds of cpu time at the cost of "fixing" 200 QString issues.  Regarding "copy on write", I only have 2 concerns - actual bugs - and inefficiencies with matrixes.

githubdoe:
Copying wavefronts

Things like rotate that make a new wavefront.  In this case you don't want to remove the old wavefront you just want a new rotated wavefront.  However the old wavefront has many things in it the the new wavefront could use.  So I copy it to make a new wavefront that has some of the features just like the old wavefront.  Here is use the word "copy" loosely. Sometimes I had to use the "=" to get what I wanted.

So you will see that sometimes to make a brand new wavefront I might use a copy or = to get a new wavefront started because a blank wavefront would then need to have a lot more operations added.

I image I wanted a full copy and not just a reference to the old mat because I'm going to allow other code to make changes and don't want to modify that old mat.

IIRC cv::Mat has an efficient form of copy to keep memory usage down where it usually gives you a reference.  It has a clone method if you really want a separate copy.  There are times especially in debug where I need plot the Mat.  However to draw it it's format needs to be changed to a different number of planes or data types.  Doing so to the original mat is not desired so then I cloned it.  

Cloning a mat can lead to a lot of storage being used up and not released.  So back in the day when DFTFringe use to run out of memory I tried to find all places where the wavefronts were created and released so that cv::Mat was actually released as appropriate.  I'm not sure I ever succeeded 100%.

atsju:
I think clazy will probably output about 1000 warnings. It's my plan also to run it (later) and maybe integrate it to github action as zero warning policy. It does much more than warn about the copy on write from my understanding

For the rest I agree Georges

githubdoe:
clazy 👍 
I think it is one of the new additions that I don't understand.  I flags all kind of naming issues that I don't plan on resolving because the names it flags were created by Qt creator itself and not me.

atsju:
> Copying wavefronts
> 
> Things like rotate that make a new wavefront. In this case you don't want to remove the old wavefront you just want a new rotated wavefront. However the old wavefront has many things in it the the new wavefront could use. So I copy it to make a new wavefront that has some of the features just like the old wavefront. Here I use the word "copy" loosely. Sometimes I had to use the "=" to get what I wanted.
> 
> So you will see that sometimes to make a brand new wavefront I might use a copy or = to get a new wavefront started because a blank wavefront would then need to have a lot more operations added.
> 
> I image I wanted a full copy and not just a reference to the old mat because I'm going to allow other code to make changes and don't want to modify that old mat.
> 
> IIRC cv::Mat has an efficient form of copy to keep memory usage down where it usually gives you a reference. It has a clone method if you really want a separate copy. There are times especially in debug where I need plot the Mat. However to draw it it's format needs to be changed to a different number of planes or data types. Doing so to the original mat is not desired so then I cloned it.
> 
> Cloning a mat can lead to a lot of storage being used up and not released. So back in the day when DFTFringe use to run out of memory I tried to find all places where the wavefronts were created and released so that cv::Mat was actually released as appropriate. I'm not sure I ever succeeded 100%.

I don't think we need to have the copy ctor do the clone. If my new understanding is correct, you will get a copy automagically when you modify the object for your drawing. I may propose the change and then you test the impact
```C++
a = b; //does a light copy with pointer
a.modify(); // clone happens magically here and a if full copy of b
```

As for the missing release and free, I think the best is to not use `new`, keep objects in a limited scope and use the default ctor and dtor when possible. Doing so you will never have leaks and objects live their life in peace.

githubdoe:
cv::Mat has it's own way of new and release.  I think it uses reference counting to decide when to free.  I had trouble getting it to free back when I was trying to free up memory.

githubdoe:
Perhaps the copy and = work as you describe.  But I can't image that the compiler is smart enough to change a reference into a clone after discovering the code would like to modify it.  Are you really sure that happens?

Sorry but I don't have much time to research things like that these days and don't plan on trying.

atsju:
> cv::Mat has it's own way of new and release. I think it uses reference counting to decide when to free. I had trouble getting it to free back when I was trying to free up memory.

Yes it's my understanding also.
Maybe it's managed better now.

atsju:
> Perhaps the copy and = work as you describe. But I can't image that the compiler is smart enough to change a reference into a clone after discovering the code would like to modify it. Are you really sure that happens?
> 
> Sorry but I don't have much time to research things like that these days and don't plan on trying.

It's not the compiler, it's as you say with a reference counter. It's definitely possible, but I'm unsure if my understanding is correct or not at this time. I discovered this today and would need to test. So my first comment about depp VS light copy was wrong because utlimately the object was deep copied when needed.
See https://www.milosolutions.com/en/blog/what-is-copy-on-write/ and https://www.youtube.com/watch?v=uZ68dX1-sVc

Don't be sorry. This code is old but works. Things have changed but I have no doubt you experienced issues during development and fixed then fine in the best way possible at the time. I do not expect you to update every line of code each time a new fancy feature is out.

For the moment I would just not touch it and focus on bug fixes and high priorities

githubdoe:
Wavefronts are contained in a list of pointers to wavefronts that is really a global but contained in mainwindow.    I can't recall any place one would pass a copy of one.  You want to work on the existing one or make a complete copy of it and then modify.

githubdoe:
By the way "Const" was a new construct with a confusing syntax based on where it was placed in the declaration of an object.  I never really ever got it figured out where to place it.  So mostly I ignored it and only used it when the compiler would complain that it could not convert a variable to the right const state.

atsju:
> Wavefronts are contained in a list of pointers to wavefronts that is really a global but contained in mainwindow. I can't recall any place one would pass a copy of one. You want to work on the existing one or make a complete copy of it and then modify.

I don't think you pass wavefronts in any function.
However, note that you used operator= on many places for wavefronts. This means that copies (light or full) are done by the default assignement operator of mat. This was disturbing my as you declared the copy ctor to do clone(). Slightly different because one does always deep copy and the other does default thing of cv::mat. => IMHO both could/should use default.

> By the way "Const" was a new construct with a confusing syntax based on where it was placed in the declaration of an object. I never really ever got it figured out where to place it. So mostly I ignored it and only used it when the compiler would complain that it could not convert a variable to the right const state.

const is simply read only. Nothing fancy but I notice it's much more used now than it was in the past. Compilers and sattic analysis tools complain more. I (and others) have asked ST to update their HAL code for 3Y now to add const where needed. It's still in progress...
`foo(const type&a)`  means the function will never touch `a`. Cool for documentation and helping the compiler. You are not required to pass const objects to such a function. However you cannot pass a const object to a function that may modify the object like `bar(type &a)` even if it doesn't modify it in reality.

gr5:
I had a similar experience with const.  It has too many meanings.  And either const and "&" came out at the same time or I learned about them around the same time so we were not using reference but instead pointers so there was this a lot:
const char const * pc = "xx"; 

Where one const refers to the pointer changing to point to something else and the other const refers to the value at that location.  Then you have functions with "const" after them and const as it applies to classes and it got confusing.  But there were certainly times when I used lots of const in many places to show intent and to reduce bugs but that was more than 20 years ago.

Until DFTF.  The last 20 years I mostly did other languages (java, javascript, php) where I didn't have to worry about pointers, references, or const very much.

And when I did use C++ the last 20 years, it was mostly with firmware in which case we were highly discouraged to use the heap for anything.  So we also didn't have to worry about things like copy constructors.

Anyway, all of the code I can currently think of should be fine as is but there may be a case where operator= default action is going to create these "weird" copy-on-write (COW) matrixes and if we pass those matrices into other functions there may be many copies being made.  Or if we reference the elements with [] instead of ".at()".  So it's worth keeping all this in mind.

90% of c++ programmers I've worked with don't even know what a copy constructor is.  And for some reason I thought operator= defaulted to use the copy constructor if operator= was not defined but the copy constructor was.  I haven't looked that up though.

atsju:
`const char const * pc = "xx";` I think you mean `const char * const pc = "xx";`
because `const char* pc = "xx";` is same as `char const * pc = "xx";`

githubdoe:
I believe I used "=" when I wanted to create a new wavefront from the old.  That got the new cv:Mat data type created to match those in the old one.  I don't remember why I did it that way instead of modifying the constructor and making an explicit copy constructor (assuming I did not make a copy constructor.)

Yes "const" has several uses depending on where it is placed and whenever I tried to use it I got in trouble with the compile objecting to me passing things into a function.

gr5:
Okay I just realized something.  First know that operator= is the assignment operator.  when you do:
wavefront wf = oldwavefront;

Instead of constructing wf and then using the assignment operator, it uses the copy constructor directly (I know that when you pass parameters it also uses the copy constructor - if you pass wf as a parameter adjust_waves(wf) would call the copy constructor for the receiving function)

Dale uses this form I believe so where I thought he was using operator= he was actually using copy constructor.

I didn't actually look at the code - this is from memory.  But I have new knowledge when looking at code in the future.

gr5:
Yes Dale, you created a copy constructor for wavefront.  You can see it if you go to the top of this page and click "files".  Although you will only see the .h file for it.

This started this whole conversation because I didn't fully understand Juliens changes to the copy constructor and assignment operator.  I did suspect (and it was true) that his changes changed nothing other than compiler warnings.  But it brought up other thoughts and concerns.

githubdoe:
George.  Like you I had a long pause in C++ programing.  Programming mostly all in python for 10 years.  Then going back to c++ just for DFTFringe.  There could be places where I passed wavefronts around thinking they would not copy when instead they did.

gr5:
If you passed a pointer to a wavefront then it's fine.  Which is mostly what you do.  For example you have an array of pointers of wavefronts (I think pointers?) and there is lots of code that has "wf->" I think.  So I think it's mostly fine.  But I'll keep it in mind.

PR #64, opened 2023-08-03
fix all remaining GCC warnings

### LINKED ISSUE #59: uninitialized variables
After cleaning over 600 warnings, I finally found 2 wich could have an effect.

https://github.com/githubdoe/DFTFringe/blob/9320be9a066892862e49a2cd4edae2935cab1efb/dftarea.cpp#L1566
this variable is used uninitialized

Same for https://github.com/githubdoe/DFTFringe/blob/9320be9a066892862e49a2cd4edae2935cab1efb/igramarea.cpp#L673

@gr5 @githubdoe I let you fix them as I still do not have the level to dig into it and you will best know the effect of this potential problem.

atsju:
With this PR, we should have 0 warnings left :)

fix #59

### DIFF
--- a/igramarea.cpp
+++ b/igramarea.cpp
@@ -670,7 +670,7 @@ void IgramArea::findCenterHole(){
     if (end > roi.cols)
         end = roi.cols;
     if (start < 0) start = 1;
-    double rmean;
+    double rmean = 0.;
     for (int rad = start; rad < end; ++rad){
         MainWindow::me->progBar->setValue(rad);
         MainWindow::me->progBar->setFormat(QString("Radius: %1").arg(rad));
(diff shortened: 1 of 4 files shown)


igramarea.cpp
```diff
@@ -2098,7 +2103,7 @@ void IgramArea::CenterOutlineActive(bool checked){
 void IgramArea::loadOutlineFile(QString fileName){
     std::ifstream file(fileName.toStdString().c_str());
 
-    long long int fsize = file.tellg();
+    std::ifstream::pos_type fsize = file.tellg();
```
gr5, line 2106: Oh yeah this line of code is weird.  I even stepped through the code to verify and fsize is always 0 as expected.  Then this zero value is subtracted off from the end of file position.  It indeed gives you the size of the file.  Which is helpful later in the code.

I think you might merge conflicts here.  Anyway I don't particularly think you should change anything.

gr5 (approved):

PR #65, opened 2023-08-04
small renaming after PR63

atsju:
I also removed the const we discussed about
#63

### DIFF
--- a/igramarea.cpp
+++ b/igramarea.cpp
@@ -1063,7 +1063,7 @@ bool IgramArea::openImage(const QString &fileName, bool showBoundary)
         qDebug() << "file exists";
         deleteRegions();
         qDebug() << "del regions";
-        loadOutlineFile(finfo.absoluteFilePath());
+        loadOutlineFileOldV6(finfo.absoluteFilePath());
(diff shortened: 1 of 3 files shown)


igramarea.cpp
```diff
@@ -2098,7 +2098,7 @@ void IgramArea::CenterOutlineActive(bool checked){
     update();
 }
 
-void IgramArea::loadJsonOutlineFile(const QString fileName){
```
gr5, line 2101: You didn't have to remove this.  It was fine.  It's telling the code inside the function that you are not allowed to modify the string.

In addition, because QString supports TOW, it prevents any copying from happening and is more efficient.  Anyway I really don't care about saving a microsecond here either way.

gr5: I mean "COW".  not "TOW".

gr5 (approved):

igramarea.cpp
atsju: Correct. I did it for homogeneity with the other functions without thinking too much about COW.
I need to read all this again to fully understand how it works.

PR #66, opened 2023-08-04
Fix all ui files warnings

atsju:
Some have simply been edited with qtdesigner.
Some needed manual edition, and I believe they would be same if they had been fully recreated with a recent QTdesigner.

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -16,7 +16,7 @@ jobs:
       - run: sudo apt install -y apt-utils build-essential wget qt5-qmake qt5-qmake-bin qt5-assistant qtbase5-dev qtmultimedia5-dev libqt5charts5 libqt5charts5-dev libqt5multimedia* libqt5datavisualization5-dev libqt5datavisualization5 libopencv-core-dev libopencv-core4.5d libopencv-dev libqwt-qt5-6 libqwt-qt5-dev libarmadillo-dev libarmadillo10
       - run: qmake
       - uses: ammaraskar/gcc-problem-matcher@master
-      #- run: echo "::add-matcher::.github/matcher/uic_matcher.json" # deactivate uic matcher for now to priorize GCC warnings
+      - run: echo "::add-matcher::.github/matcher/uic_matcher.json"
       - run: make -j4
-      #- run: echo "::remove-matcher owner=uic-problem-matcher::" # deactivate uic matcher for now to priorize GCC warnings
+      - run: echo "::remove-matcher owner=uic-problem-matcher::"
       
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -355,9 +355,9 @@ jobs:
           (Get-Content DFTFringe/DFTFringeInstaller/packages/com.githubdoe.DFTFringe/meta/package.xml).replace('MY_AUTOMATED_DATE_STRING', '${{env.NOW}}') | Set-Content DFTFringe/DFTFringeInstaller/packages/com.githubdoe.DFTFringe/meta/package.xml
           
       - run: cd DFTFringe ; ..\${{env.QT_version}}\mingw81_64\bin\qmake.exe
-      #- run: echo "::add-matcher::.github/matcher/uic_matcher.json" # deactivate uic matcher for now to priorize GCC warnings
+      - run: echo "::add-matcher::.github/matcher/uic_matcher.json"
       - run: cd DFTFringe ; mingw32-make -j4
-      #- run: echo "::remove-matcher owner=uic-problem-matcher::" # deactivate uic matcher for now to priorize GCC warnings
+      - run: echo "::remove-matcher owner=uic-problem-matcher::"
 
       - name: copy needed dlls not found by windeployqt
         run: |
--- a/astigstatsdlg.ui
+++ b/astigstatsdlg.ui
@@ -30,138 +30,130 @@
      </property>
      <layout class="QHBoxLayout" name="horizontalLayout_2">
       <item>
-       <widget class="QPushButton" name="clearPb">
-        <property name="text">
-         <string>clear</string>
-        </property>
-       </widget>
-      </item>
-      <item>
-       <spacer name="horizontalSpacer_4">
-        <property name="orientation">
-         <enum>Qt::Horizontal</enum>
-        </property>
-        <property name="sizeHint" stdset="0">
-         <size>
-          <width>3</width>
-          <height>20</height>
-         </size>
-        </property>
-       </spacer>
-      </item>
-      <item>
-       <widget class="QPushButton" name="zernikePB">
-        <property name="toolTip">
-         <string>Load zernike values for the plot from a .csv file.</string>
-        </property>
-        <property name="text">
-         <string>Add Zernike csv</string>
-        </property>
-       </widget>
-      </item>
-      <item>
-       <spacer name="horizontalSpacer">
-        <property name="orientation">
-         <enum>Qt::Horizontal</enum>
-        </property>
-        <property name="sizeHint" stdset="0">
-         <size>
-          <width>2</width>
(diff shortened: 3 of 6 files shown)


githubdoe (changes requested):
It is useless to modify .ui files.  They will change the next time Qt is compiled using the Qt  build which is what I do.

atsju:
Could you explain me what is this QT build ? 
because if they change it means they are not input files and something is on your computer but not on github.
Or are they opened, modified, and saved by QT build for obscure reasons ?

githubdoe (commented):
It is useless to modify .ui files.  They will change the next time Qt is compiled using the Qt  build which is what I do.

githubdoe:
Damn I hate .Arg().   It is so hard to figure out what is being printed. After one has used PRINTF syntax  I can see in one glance what is being printed.  With .arg() you have to look in two places and then remember what the syntax is for arg and what each parameter may be specifying.  Aaaaaaaaggggg.

githubdoe (approved):
I will approve but just know the any and probably all .ui files will change the first time I compile.  I don't think you want to keep trying to change them.

githubdoe (commented):
It is useless to modify .ui files.  They will change the next time Qt is compiled using the Qt  build which is what I do.

githubdoe:
I just realized that perhaps you did not edit the .ui files that they are changing just because of re compliling.

atsju:
> I will approve but just know the any and probably all .ui files will change the first time I compile. I don't think you want to keep trying to change them.
Please explain why .ui would be modified by compilation? this makes NO sense for me.
More probable is you open them with an IDE that modified them without telling you. But not the compilation process. Yes the simple fact opening .ui with QTdesigner and saving did some unexpected modifications but it's not the compilation process.

> I just realized that perhaps you did not edit the .ui files that they are changing just because of re compliling.
No I did it on purpose and manually directly in file and/or with QTdesigner

githubdoe:
Qt creates the .ui files from data provided by QtCreator which is derived from the GUI that lets us create Qt forms (windows, dialogs and such).

Build is Qt lingo for make the files necessary to compile the program.
 It will create .cpp and .h files from those .ui files.  Those cpp files with "ui" in the name.

If one modifies the from like adding a button or deleting something or moving something it will recreate the .ui file.  
The next time you compile it will then make or remake the  .cpp and .h files related to that .ui file.

If I do a clean build it will delete all .cpp and .h files derived from those .ui files as well.  I will always do a clean build just before creating a final release.  Sometime due to a bug in Qt creator one needs to delete those related .cpp and .h files to get them updated.  So in the Qt IDE you can say clean.  It will delete all those files so that they can then be remade during the nest "Build"

githubdoe:
Yes the .ui files will stay until one uses the GUI editor and not when they are compiled.    However Qt designer owns them and I think they should not be modified by the developer.

atsju:
Thank you Dale I think it's clearer for me now.

I don't think QTdesigner owns the file. 
The .ui is just the way QTdesigner stored the thing you just draw.
image

From my understanding when you modify a windows, the QTdesigner opens the .ui you can modify the visual and it stores the .ui back. If QTdesigner "owned" the thing then you would have it stored somewhere in the project and it's note the case.

I agree .ui shall not be manually modified but only using QTdesigner or creator. Here I did only small modifications that do not break the compatibility with corresponding .cpp and .h.

If you "clean build" a .ui from now because you need to add a button or something, I hope that the warning will not come back. Because the IDE does not create the .ui from nothing or owning, it just modifies existing file. At least that's my understanding right now. I might be wrong.

githubdoe (commented):
Sorry I still am not use to Git hub code reviews.  When I do diffs in Qt using git tools I ignore changes to .ui or at least remember changing the gui and thus the .ui changed.

Here seeing them on Github the triggered a false worry that you were changing them manually and not through the Qt IDE.

atsju:
Please do not merge without testing. I did not have time to test the build yet

gr5:
Dale I think you confused *.ui files with ui_*.h files.  The .h files are created based on the .ui files.  We don't check in the ui_*.h files as they are created as part of the build process.

The *.ui files however can be edited by opening them in qt creator and changing things from that editor which is what Julien did I think.  For example you had several layout widgets with the same name and I guess that somehow created compiler warnings.  In QT Creator you can change the name over on the properties window (like you can change many other things).

Even if he edited the ui files directly, they should hopefully still open in qt creator.  I'll test that today.

githubdoe:
There are things you can do that corrupt Qt's ability to find or create slots based on the objects name.  I don't know exactly how it happens but it did to me even without my editing the .ui file.  I have a couple of .ui's that are now useless if I ever want to add objects who could send signals.  Very frustrating.  Sometimes I even recreated the whole form over again.  That makes me worry about manually modifying the .ui file.  Especially changing names of objects.  But who knows.

gr5:
Okay so I can imagine if you change the name, now the name of the slot or signal could change.  So I hear your concern.

But I think he only edited the names of layouts which won't have any signals used.  I'm going to look at this some more.  I've already opened the first ui file in qt creator - in the gui editor.  So far it seems fine.

He also moved some things around...

atsju:
I made some tests on the app and it seems to work fine.

Also I thank you for the explantations Dale. I better understand your concerns now. I think what you describe is not supposed to happen but this must be frustrating when it does. And from your experience : it does. 
I think that part of your experience is why clazy [level0 checks](https://github.com/KDE/clazy/blob/master/docs/checks/README-connect-by-name.md) recommends to not use [connect-by-name](https://github.com/KDE/clazy/blob/master/docs/checks/README-connect-by-name.md)

> He also moved some things around...

Only the 4 `horizontalSpacer_4` have been reordered (but I can put them back in original order). Remember to use the "hide whitespaces" option during github review. Changes have been surgical.

githubdoe:
Sorry but I don't understand the issue with connect by name.  I'm use to it and expect it and don't know how to do otherwise.  When I click on those links it does not say much of anything except that people don't like it.  Clicking on the recommend link to learn more there is no more info other than to tell you that Qt uses connect by name.  Hmmm.

So there are hundreds of names with on_...   I like them because then I know what to look for and know that the routine is providing a service for a gui object.  Yes I know that changing the name of the GUI object will beak the connection and know where and how to fix and recover the intended connection.

Concerning connects:
  Qt introduced a newer connect macro system but I can never remember it's syntax and when I try to use it or look it up I get confused.  So I did not use it often.  I understand why the older system is not as good since it only error checks at run time.  But I can never remember how to do the other connect.  Not against using it however.  

Well I can never remember the old connect system either in truth.  I just go find old code I wrote to remind me if and when I need to do a connect.  Since there is little old code that uses the new system it is hit or miss of me finding an example.  Mostly miss.

atsju:
> Sorry but I don't understand the issue with connect by name. I'm use to it and expect it and don't know how to do otherwise. When I click on those links it does not say much of anything except that people don't like it. Clicking on the recommend link to learn more there is no more info other than to tell you that Qt uses connect by name. Hmmm.

Same for me. I don't plan to change anything yet. I will keep you posted if I find the alternative method.

In the meantime I will just add an action for using clazy without this one warning. Then we see what we change or not

gr5:
I opened the 4 .ui files in the design editor in qt creator and was able to move components around and such.  It all seems fine.  I did not save any changes to the ui files of course.

Then I built and ran it and used the 4 related windows. They are all fine.

So editing is still working fine.  Functionality is still working fine.  So I'm going to merge now.

gr5:
By the way, the only build errors left seem to be in armadillo so new warnings will *really* stand out now.  I understand that these are also fixed in a newer version of armadillo but there's really not very many left.

atsju:
I just see now that I missed these 2
image

Also about the left warnings:
- armadillo has already been fixed, no more warning from there
- those about use-alpha are only on windows and it's a know compiler bug because it's a (very) old compiler version
- 2 are left in qwt

PR #68, opened 2023-08-05
run with clazy on linux

### LINKED ISSUE #67: clazy
I just managed to run clazy with exclusion pattern to get only warnings from DFTF code.

Total 8697 warnings
But 8512 of them are [connect-by-name](https://github.com/KDE/clazy/blob/master/docs/checks/README-connect-by-name.md) and I suppose we must ignore them because nobody plans on updating this

This means we have 185 more meaningfull warns. Not too bad.

clang also generated some `[-Woverloaded-virtual]` and `[-Wabsolute-value]`. I will probably fix the `[-Wabsolute-value]`

atsju:
OK I tried to add [clazy](https://github.com/KDE/clazy#list-of-checks) to the build

all level 1 checks 
ignore clazy-no-connect-by-name

~~I don't think we should merge this one as long as all warnings from the tool are not removed. So I will just keep it in draft.
Because github can only display 10 warnings total then other more important warnings from gcc itself won't stick out.~~
I think we can let it run but remove the problem matcher that displays the warnings to github.

close #67

### DIFF
--- a/.github/workflows/build-linux-clazy.yml
+++ b/.github/workflows/build-linux-clazy.yml
@@ -15,7 +15,7 @@ jobs:
       - run: sudo apt update
       - run: sudo apt install -y clazy apt-utils build-essential wget qt5-qmake qt5-qmake-bin qt5-assistant qtbase5-dev qtmultimedia5-dev libqt5charts5 libqt5charts5-dev libqt5multimedia* libqt5datavisualization5-dev libqt5datavisualization5 libopencv-core-dev libopencv-core4.5d libopencv-dev libqwt-qt5-6 libqwt-qt5-dev libarmadillo-dev libarmadillo10
       - run: qmake -spec linux-clang QMAKE_CXX="clazy"
-      - uses: ammaraskar/gcc-problem-matcher@master
+      #- uses: ammaraskar/gcc-problem-matcher@master  #TODO 2023/08/05 re-enable problem matcher when all warnings are fixed
       # ignore noisy dirs from QT files itself
       # all level 1 checks but ignore clazy-no-connect-by-name
       # no parallel make with clazy to not mess log
--- a/.github/workflows/build-linux-clazy.yml
+++ b/.github/workflows/build-linux-clazy.yml
@@ -1,0 +1,22 @@
+name: build-linux-clazy
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  workflow_dispatch:
+  workflow_call:
+
+jobs:
+  build-linux-clazy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3 
+      - run: sudo apt update
+      - run: sudo apt install -y clazy apt-utils build-essential wget qt5-qmake qt5-qmake-bin qt5-assistant qtbase5-dev qtmultimedia5-dev libqt5charts5 libqt5charts5-dev libqt5multimedia* libqt5datavisualization5-dev libqt5datavisualization5 libopencv-core-dev libopencv-core4.5d libopencv-dev libqwt-qt5-6 libqwt-qt5-dev libarmadillo-dev libarmadillo10
+      - run: qmake -spec linux-clang QMAKE_CXX="clazy"
+      - uses: ammaraskar/gcc-problem-matcher@master
+      # ignore noisy dirs from QT files itself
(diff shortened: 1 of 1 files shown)


gr5:
Where would one go to see the output of clazy?

gr5:
Oh.  I found it.  Here is one of the very thing we were talking about:
astigstatsdlg.cpp:262:5: warning: allocating an unneeded temporary container [-Wclazy-container-anti-pattern]
    foreach(QString name, groups.keys()){

atsju:
> Where would one go to see the output of clazy?

directly in the action run log but without highlighting
https://github.com/githubdoe/DFTFringe/actions/runs/5770975905/job/15644494054?pr=68
(you can also with in raw format https://pipelines.actions.githubusercontent.com/serviceHosts/650bc925-ae05-4f18-b40d-173cd102dab0/_apis/pipelines/1/runs/239/signedlogcontent/2?urlExpires=2023-08-05T14%3A26%3A20.9183486Z&urlSigningMethod=HMACV1&urlSignature=3mSkJQnBiTsxNpZjyrS1rmYStEdp%2B3gExegP32J4kgI%3D)

gr5:
GOOD NEWS:
The wasted memory allocation we were talking about only seems to happen in 7 lines of code and I think it's fixed by simply adding "const" and they are mostly QStrings and I think one class Double.  So not a lot of wasted memory or cpu time (no matrixes).

And Julien - there's no need to fix all the warnings.  We don't need to get the clazy warnings down to zero please.  But we could fix a few of them.  

The memory ones are fixed probably with simple addition of const for each one.  Then the compiler will know that it won't be modified and won't create a copy.

The "unused variable" ones are simple as well - just delete them.  And they should be removed I suppose.

But there will always be lots of clazy warnings so let's just fix the ones that improve code, fix bugs, or improve readability.

For example the abs() ones that warn that it's an integer function.  Dale probably knew that so look the code over carefully before deciding if it should be changed to fabs() or std::abs() etc.  It can probably be left as is.  But looking the code over carefully to understand what is going on exactly - and deciding if this is a bug - that is worth doing.

atsju:
I'm already on the unused variables. those are easy.
I looked at [-Wclazy-qstring-arg] also and for these I decide to not change the code but add an ignore comment. Because readability takes precedance over extra small efficiency improvement.
All others, I will look at case by case basis and I target 0. I don't think it can hurt but of course I will try to understand before doing any change.

About abs, if `abs` was really meant to be used instead of `fabs` code should be rewritten to make an explicit rounding. I really do not expect it but you are right so I will check the code and keep you updated.

gr5 (approved):

gr5:
why is this still "in progress"?

atsju:
> why is this still "in progress"?

You told me `abs` could be integer on purpose. Now I need to read the code and check if it's possible that using float messes up the thing

Note std::abs will work with float and int whenever needed but I still want to make sure that Dale didn't need an int rounding.

atsju:
Unrelated : I'm currently playing with HeNe laser, gratting and trying to figure out the Laser diode wavelength... What should be simple math is not working. I will come back to the code after.

gr5:
No this PR just adds clazy.  I'm ready to pull.

The abs() thing is in the other PR.  I just happened to mention abs() in this PR because I happened to notice it in the clazy output.  I didn't realize you had another PR related to abs() until after I wrote the above comments.

atsju:
OK fixed. This one can be merged indeed

atsju:
> The memory ones are fixed probably with simple addition of const for each one. Then the compiler will know that it won't be modified and won't create a copy.

FYI I tried and it's not sufficient. There is extensive use of QMap and one must use [const iterator](https://doc.qt.io/qt-6/qmap-iterator.html) with range-for loop. I will share the fixes when ready.

Also note that QT foreach is not recommended anymore. On must use range-for loops from C++11.
Also It's recommended to use std:: containers instead of QT container (Qmap, Qvector, ...). Except when interfacing with QT features that require such containers of course.
They (foreach and containers) are here for historical reasons because QT existed on multiple platforms long before C++ had these cool features. 
I do not recommend to update existing/working code, however new features should take this into account.

gr5:
Yeah I thought it was strange that this QT employee was giving a talk saying "don't use QT containers, use std" and then I realized they are stuck and can't remove them from their library because then lots of customers will complain (including me!).

PR #71, opened 2023-08-08
fixClazy warnings

atsju:
fix 6 occurences of `[-Wclazy-container-anti-pattern]`
I removed foreach loop each time because it's not recommended anyway anymore.

I tested the impact on the different windows except "artificial null error margin" (nullvariationdlg) as I did not understand how to use it.  I have seen no obvious effect.

### DIFF
--- a/astigstatsdlg.cpp
+++ b/astigstatsdlg.cpp
@@ -258,15 +258,16 @@ void astigStatsDlg::plot(){
     double ymax = -1000;
     int colorndx = 0;
     QMap<QString, QPointF> means;
-    foreach(QString name, groups.keys()){
-        int size = groups[name].size();
+    for(auto it = groups.cbegin() ; it != groups.cend() ; it++){
+        QString name = it.key();
+        const int size = it.value().size();
         double *xAstig = new double[size];
         double *yAstig = new double[size];
         RunningStat xstats, ystats;
 
         QColor color = QColor(Qt::GlobalColor( 7 + colorndx%12 ));
         QVector<QPointF> points;
-        foreach (measure data, groups[name]){
+        foreach (const measure &data, it.value()){
             xAstig[i] = data.p.x();
             points << QPointF(data.p.x(), data.p.y());
             xstats.Push(xAstig[i]);
@@ -389,11 +390,9 @@ void astigStatsDlg::plot(){
     if (ui->bestFitCB->isChecked()){
         double *xvals = new double[means.size()];
         double *yvals = new double[means.size()];
-        int i = 0;
-        foreach(QString name, means.keys()){
-            xvals[i] = means[name].x();
-            yvals[i] = means[name].y();
-            ++i;
+        for(const QPointF &value: qAsConst(means)){
+            xvals[i] = value.x();
+            yvals[i] = value.y();
         }
 
         CircleData d(means.size(), xvals, yvals);
@@ -799,10 +798,10 @@ void astigStatsDlg::on_distribution_clicked()
         html.append("<p><img src='" +imageName + "'/></p>");
     }
     // for each group
-    foreach(QString Name, groups.keys()){
-
-        QList<QPointF> items = groups[Name];
-        int size = items.size();
+    for(auto it=groups.cbegin() ; it!=groups.cend() ; it++){
+        const QList<QPointF> items = it.value();
+        const QString Name = it.key();
+        const int size = items.size();
         float *xAstig = new float[size];
         float *yAstig = new float[size];
         for (int i = 0; i < size; ++i){
--- a/astigstatsdlg.cpp
+++ b/astigstatsdlg.cpp
@@ -390,9 +390,11 @@ void astigStatsDlg::plot(){
     if (ui->bestFitCB->isChecked()){
         double *xvals = new double[means.size()];
         double *yvals = new double[means.size()];
+        int i = 0;
         for(const QPointF &value: qAsConst(means)){
             xvals[i] = value.x();
             yvals[i] = value.y();
+            ++i;
         }
 
         CircleData d(means.size(), xvals, yvals);
--- a/nullvariationdlg.cpp
+++ b/nullvariationdlg.cpp
@@ -117,22 +117,23 @@ void nullVariationDlg::on_diameterDelta_valueChanged(double val)
     set.setValue("nullDiamTol", ((isMm) ? 1. : 25.4) * val);
     calculate();
 }
+
 #include <qmap.h>
 #include <qwt_plot_histogram.h>
+
 QMap<double,int > histo(const QList<double> data, int bins){
     auto minma = std::minmax_element( data.begin(), data.end());
     double range = *minma.second - *minma.first;
     double histInterval = range/bins;
     QMap<double,int> intervals;
-    for (int i = 0; i < bins; ++i)
+    for (int i = 1; i <= bins; ++i)
     {
-        intervals[ *minma.first + histInterval * (i+1)] = 0;
+        intervals[ *minma.first + histInterval * i] = 0;
     }
-    QVector<double> keys = intervals.keys().toVector();
     for (int i = 0; i < data.size(); ++i){
-        for (int j = 0; j < bins; ++j){
-            if (data[i]<= keys[j]) {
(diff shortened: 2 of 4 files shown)


githubdoe:
Well that makes me a little sad.  There was a time when Qt said to prefer their containers and for loops to those of the standard libraries.  Times change.  However if you try to keep up I fear you will just spend a lot of time without any improvement as far as the user is concerned.  Those are changes just for changes sake.  Go ahead we have a different maintenance philosophy however.

atsju:
> Well that makes me a little sad. There was a time when Qt said to prefer their containers and for loops to those of the standard libraries. Times change. However if you try to keep up I fear you will just spend a lot of time without any improvement as far as the user is concerned. Those are changes just for changes sake. Go ahead we have a different maintenance philosophy however.

I will mitigate this just a little bit. I do not plan to change every foreach and every container to std::
These occurrences "needed" changes because the tool detected a small performance issue so I changed them to for loops in the process. 
For me removing warnings is a good learning experience and I do not feel like I can do anything complex for the end user anyway. Drawback is of course your review time. I believe this will help developer to have a 0 warning policy so in the long term it will still limit bug and help the end user.

atsju:
@githubdoe I'm working on some other simple cleanups that I will share once this is merged and I have 2 questions: 

1. `mikespsirinterface.cpp` and `mikespsirinterface.h` are unused. I'm already removing them from build. Can I delete the file or do you want to keep them for historical reasons ? (I do not recommend to keep them but if we do I would rename them to .cpp_ and .h_ or move them to different folder)
2. Please have a look at `zernEnableUpdateTime`in the code. It seems unused. Can you confirm it can be removed and there is no unknown (to me) side effect ?

githubdoe:
They must be kept.  That is the original source that Mike Peck provided to do PSI analysis.

atsju:
@githubdoe anything to add about `zernEnableUpdateTime` ?

githubdoe:
Ok I found it.  Yes it can be removed.  Probably was meant to help set a timer that would trigger to update a display but either never got implemented or was no longer needed.

githubdoe (commented):
That is code I hate to review.  I don't have the time or desire to do when it was code that already worked and did not need changing.  Sorry.  I'm just not any good at reading code to find bugs.  So I can not say if the code does what we expect.

atsju:
> That is code I hate to review. I don't have the time or desire to do when it was code that already worked and did not need changing. Sorry. I'm just not any good at reading code to find bugs. So I can not say if the code does what we expect.

I agree it's not directly usefull to change code that worked but still, I spotted some dead code, file compiling while unused and minor (non-)issues. The tool even spotted some `emit` on non-signals that I will fix. 
I'm sincerely sorry to make you endure this but it's necessary for maintainability in long term. I expect this tool to stay for the next 10Y at least and it will necessarily get some changes with the OS and QT evolution. See this as an investment.
So far we did not even spot any regression from my "maintainability" changes, which makes me little anxious TBH :sweat_smile:

astigstatsdlg.cpp
```diff
@@ -389,11 +390,9 @@ void astigStatsDlg::plot(){
...
         double *xvals = new double[means.size()];
         double *yvals = new double[means.size()];
-        int i = 0;
-        foreach(QString name, means.keys()){
-            xvals[i] = means[name].x();
-            yvals[i] = means[name].y();
-            ++i;
+        for(const QPointF &value: qAsConst(means)){
+            xvals[i] = value.x();
+            yvals[i] = value.y();
```
atsju, line 395: https://github.com/githubdoe/DFTFringe/pull/72#discussion_r1290459126

What happened to updating i?

astigstatsdlg.cpp
```diff
@@ -258,15 +258,16 @@ void astigStatsDlg::plot(){
     double ymax = -1000;
     int colorndx = 0;
     QMap<QString, QPointF> means;
-    foreach(QString name, groups.keys()){
```
gr5, line 261: so isn't this fixed by just adding a const like this?
`foreach(const QString name, groups.keys()){`

atsju: No. The "problem" is that keys() generated a new list containing the keys. Which is "inefficient".

nullvariationdlg.cpp
```diff
@@ -117,22 +117,23 @@ void nullVariationDlg::on_diameterDelta_valueChanged(double val)
...
 #include <qmap.h>
 #include <qwt_plot_histogram.h>
+
 QMap<double,int > histo(const QList<double> data, int bins){
     auto minma = std::minmax_element( data.begin(), data.end());
     double range = *minma.second - *minma.first;
     double histInterval = range/bins;
     QMap<double,int> intervals;
-    for (int i = 0; i < bins; ++i)
+    for (int i = 1; i <= bins; ++i)
```
gr5, line 129: I think the old code with i=0 was fine.  The new code is also fine.  I have no preference so in general we shouldn't change this.  But it's fine.

atsju: Yes I agree with you and try to restrain myself for every little changes I see. However here there was a `i+1` in each loop. 
As we are used to `=0;<n` and to `=1;<=n` I preferred the latest in this case. I know it's only style and I should just have let it alone...

atsju:
About the `i` update that Dale correctly spotted in this PR, I did run the code again and did not see the impact. Now I'm wondering why the wrong code is also working ... I will be back on monday only, have a nice WE.

image

githubdoe:
I agree with George.  I think it is very risky to modify all of that code for no good reason.  I'm especially concerned when I just now discover from the conversations here that some reasons for change was efficiency reasons as indicated by a warning.  Stop doing any of those please.

githubdoe:
I saw the adding "const" in various places and don't understand the need.  Please explain.

githubdoe:
IIRC The astig plots computation methods change and depends on the number of rotation angles used.  That is because it wants to draw a circle through the averages.  However if only 2 rotations were used the method of finding a circle is different than if there are more rotations.   So you need different number of rotations to full test it.

githubdoe:
The keys thing may be inefficient but it does not matter.  It is only used a few times for a few data points and is not part of a huge loop.  Don't worry about those.

githubdoe:
Never worry about efficiencies.  You likely will not make the right choices until you run a performance analyzer to discover where they matter.  However I was never able to get the performance analyzer to work with the Qt IDE.  I don't remember why.

githubdoe:
Now I remember .  Performance can only be run on a Linux kernel.

atsju:
> I saw the adding "const" in various places and don't understand the need. Please explain.

Every variable that is created and only accessed for read but never write again should be declared `const` by default. It helps the co-developper reading code AND the compiler. You might think that the compiler will optimize when he sees that variable is never write accessed but this assumption is wrong for QT containers which will detach on read access and create performance issue. Moreover, adding a const will never cause any issue. If compilation is success code might or not be better optimized. And If a write access was needed then compiler will complain. Also some places use `const int &data` which is a constant reference. Meaning data can not be used for write access (even when original variable is not const).

> IIRC The astig plots computation methods change and depends on the number of rotation angles used. That is because it wants to draw a circle through the averages. However if only 2 rotations were used the method of finding a circle is different than if there are more rotations. So you need different number of rotations to full test it.

Thanks for feedback. So I did some more tests with rotation and still did see no difference. I will push investigation further because I want to understand.

> The keys thing may be inefficient but it does not matter. It is only used a few times for a few data points and is not part of a huge loop. Don't worry about those.

I agree it's not a huge impact for those. However the warning was accurate and I had choice between ignore/mute warning or change the code. The modifications being cheap surgical changes reviewed and tested I choose this path. If you prefer me to mute those I can let the code as it is. Note I will do a detailed explantation for each change in the other PR (#72). Kindly let me know there which changes you would like to keep and which you would like me to drop.

> Never worry about efficiencies. You likely will not make the right choices until you run a performance analyzer to discover where they matter.

I agree and this was the plan. I was not going to do anything blind without analysis. Maybe I won't be able to get the analyser working, maybe I will find your code has no major performance problem and maybe I will find something. In all case you will get last word for merge and I will value the learning as I never done that in this context before. I'm more used to chase nano-amps of power consumption in MCUs.

Issue #22, opened 2023-03-26
[Feature request] crash report for people to help report


atsju:
I had 2 crashes while using official version 6.2

I have been unable to reproduce the crashes reliably so I can't even tell you what to fix. I think DFTF should catch all exceptions and either write into a log file or show a crash window for people to copy paste the call stack into an issue. It should not close/crash silently. This is standard practice to help developpers fix issues happening on "customer" side.

here are information from windows event report. Not very usefull.
```
Nom de l’application défaillante DFTFringe.exe, version : 0.0.0.0, horodatage : 0x63db083b
Nom du module défaillant : libgcc_s_seh-1.dll, version : 0.0.0.0, horodatage : 0x00000000
Code d’exception : 0x40000015
Décalage d’erreur : 0x000000000000d6a5
ID du processus défaillant : 0x3b8c
Heure de début de l’application défaillante : 0x01d95ff7eda6c2ff
Chemin d’accès de l’application défaillante : C:\DFTFringe64\DFTFringe.exe
Chemin d’accès du module défaillant: C:\DFTFringe64\libgcc_s_seh-1.dll
ID de rapport : 18c77f35-82bc-4ac2-8d20-1c7e59baf140
Nom complet du package défaillant : 
ID de l’application relative au package défaillant : 
```
```
Nom de l’application défaillante DFTFringe.exe, version : 0.0.0.0, horodatage : 0x63db083b
Nom du module défaillant : libopencv_core3412.dll, version : 0.0.0.0, horodatage : 0x5fddbfdb
Code d’exception : 0x40000015
Décalage d’erreur : 0x000000000023e078
ID du processus défaillant : 0x39d8
Heure de début de l’application défaillante : 0x01d95ff0660379b8
Chemin d’accès de l’application défaillante : C:\DFTFringe64\DFTFringe.exe
Chemin d’accès du module défaillant: C:\DFTFringe64\libopencv_core3412.dll
ID de rapport : a8ff05b6-894e-4a0f-b156-5b5ec77b43d4
Nom complet du package défaillant : 
ID de l’application relative au package défaillant : 
```

gr5:
I was thinking a log file.  Currently we write to qDebug() output which only runs if you compile debug version and that output goes to some special console (not stdout).  Instead I'd like some kind of a log file.  Crash info could be dumped into the same file.  It should include dates and times and a reliable line at the start whenever you relaunch the program.

[...]

PR #72, opened 2023-08-10
fix almost every clazy warnings

atsju:
Only some `[-Wclazy-incorrect-emit]` are left.

This should be merged only after #71 and a rebase. Thus it's draft.

Some of these clazy warnings were actually quite good on spotting some QT suboptimal uses. And clang also spotted some function override that were harmless here (of course, because it worked) but could have been pain to debug. 
This was more interesting and usefull regarding QT than all the GCC warnings I fixed before.


astigstatsdlg.cpp
```diff
@@ -389,11 +390,9 @@ void astigStatsDlg::plot(){
...
         double *xvals = new double[means.size()];
         double *yvals = new double[means.size()];
-        int i = 0;
-        foreach(QString name, means.keys()){
-            xvals[i] = means[name].x();
-            yvals[i] = means[name].y();
-            ++i;
+        for(const QPointF &value: qAsConst(means)){
+            xvals[i] = value.x();
+            yvals[i] = value.y();
```
githubdoe, line 395: What happened to updating i?

atsju: see lines 392 and 396 just above. `i` was in fact unused so I removed it.

githubdoe: I is not unused.  It is used in 394 and 395 as an index into xvals isn't it?

atsju: That's so dumb from me.
And this is why we have code review. Thanks a lot I will fix it. And also check why I didn't detect it during my test...

atsju: I fixed it in the other PR as it's part of there and not here. This one PR is not ready to merge as long as the other is not done. Rebasing will fix here also.

githubdoe: I noticed it only because that was  one disadvantage of using foreach.  You lose the index of the object if you need it.

gr5:
Julien I specifically said I didn't want to fix all the inefficiencies with short (less than 1000 character) qstrings.

Every change we make can introduce a new bug that might not be discovered until well after many people have it installed and are using it.

The only reason to do these qstring changes is if we can remove all the clazy warnings but my impression was that there are many hundreds and we are pretty much guaranteed to introduce a bug if we try to fix them all.  If there truly are only about 10 clazy warnings then I'm open to that but I suspect some of them are tricky to fix.

I'm quite happy to get rid of unused code or at least keep it from being compiled in the case of "mikes psi r" code by renaming it to something other than .cpp.

I'd much rather see you implementing the logging usse or one of the other issues you created versus fixing clazy warnings.

Instead I see clazy as possibly finding serious inefficiencies with larger blocks of memory such as the many matrixes.  I believe it didn't find any.

I could be wrong - there might only be 20 clazy warnings and in that case it's probably worth it having this new tool that can detect inefficiencies, bugs, and just ugly code.

I certainly love all the unused variables it found that gcc missed.  This improves code readability with zero chance of introducing a bug.

All the sprintf() changes were scary.  I did find one bug after the fact (already fixed in my latest pull) where one %s didn't get changed to %1.  But it was worth it because I was in the habit of ignoring warnings (not any more!).

Maybe the clazy "fixes" will be worth it but I think there are too many when I glanced at them before so I suspect not.

gr5:
Also can't we fix all these qstring issues by adding "const" and not have to rewrite as a different format of for loop?

I suspect the only reason I think it's recommended not to use for() is because some users don't know to use const.

I'd MUCH rather fix these by adding a const as that reduces the chance of introducing a bug to almost zero.

gr5:
Oh I just realized I put all these notes on the wrong pull!

atsju:
> Also can't we fix all these qstring issues by adding "const" and not have to rewrite as a different format of for loop?

I tried and it didn't work that easy.

> Oh I just realized I put all these notes on the wrong pull!

Yes thats not a big issue. You may have seen that this PR 2 first commit are from the other opened PR. I separated them because those are more complex changes than all the small renamings I did here.

> Maybe the clazy "fixes" will be worth it but I think there are too many when I glanced at them before so I suspect not.

Many of these warnings were duplicate. I swear that in current state of this PR there are only 10 "incorrect-emit" left to fix. I will do them in a different PR because it seems a bit more complex. I don't think that doing `emit` on something not a signal is a warning we can ignore.
I also see that the problem matcher is not matching well because the format is slightly different form GCC warnings. I will fix it. 
You can have a look at raw log and search for `[-W` (note that 89 of the 99 remaining warnings are identical and not from our code). For me it's worse it because we will refuse futur development introducing warnings.

> I'd much rather see you implementing the logging usse or one of the other issues you created versus fixing clazy warnings.

I was more attracted by optimizing performance with openCV or searching for memory leaks (for example I have seen some QT objects not having a parent. Of course we can argue about real VS false memory leak as we already did but it's supposed to have parents.)
But I can work on this logger if you think it's a better priority. => we need to decide on which logger to use in #22 (in parallel of keeping Qdebug).

gr5:
> Many of these warnings were duplicate. I swear that in current state of this PR there are only 10 "incorrect-emit" left to fix.
Only 10 after these.  Okay then that's fine.



> I was more attracted by optimizing performance with openCV or searching for memory leaks (for example I have seen some QT objects not having a parent. Of course we can argue about real VS false memory leak as we already did but it's supposed to have parents.) But I can work on this logger if you think it's a better priority. => we need to decide on which logger to use in #22 (in parallel of keeping Qdebug).

memory leaks are a great thing to work on.  And possible memory corruption as well.  Dale seems to think that DFTF slowly uses more and more memory the more you use it - sounds like a memory leak to me.  Don't work on log if it's not interesting to you.  Please work on things that are both interesting to you and improves the experience for users.

QT Creator crashes for me sometimes and for dale also.  For other reasons I suspect there are some potential memory corruption issues where for example writing beyond the end of a matrix corrupts something else on the heap and causes a crash or other serious (but intermittent) bug.  I think one of those has been fixed related to regions being outside of the matrix.  But there may be more.

Optimizing performance would be nice.  I have seen many potential improvements but they all make the code more complicated.

For example somewhere I saw pow(x,2) style of code instead of x*x.  pow() function probably uses a taylor series and requires many multiplications, divisions, summations.  x*x is significantly faster when applied to 47 zernike terms X 1 million "pixels".  And many other potential speedups (x*x is not more complicated but some of my other ideas are).

I'd love to see processing of igrams fast enough to process 10 frames per second.

atsju:
I @githubdoe I understand your concerns and highly value your feedback. Let me explain the changes and you will of course have last word if the changes can/should be merged or shall be dropped. I'm grown up and if you tell you refuse some of these changes I will just close the PR. I don't think you will dismiss ALL changes after my explanation, let's see.

You can click on each commit to see the actual changes. 

1. 5148f42 + 665b6a4 : `[-Wclazy-container-anti-pattern]`
Those are the changes from the other PR. Rationale is that the loop was creating a new list of keys of creating new objects that could be accessed faster with iterators or const references. Warning is valid, but impact is small when number of elements is limited.
2. acf813d : `delete 2 unused signals and rename 1 [-Wclazy-overloaded-signal]`
Rationale is `selected` and `load` are keywords from QT signals. Those functions were overriding QT signal which could cause problems. I deleted the unused signals and renamed the other offending.
3. 21522d4 : `[-Wclazy-fully-qualified-moc-types]`
Rationale explained by [clazy documentation](https://github.com/KDE/clazy/blob/master/docs/checks/README-fully-qualified-moc-types.md). Non fully-qualified typed can cause signals to break. Here it is not a signal so it has no impact. Still good practice to fully qualify.
4. f6956ed : `mute specific [-Wclazy-qstring-arg] for readability`
Using `.arg().arg()` makes 2 allocations where for int you have a simplified arg taking 2 parameters. I choose to mute the warning to keep readability instead of performance.
5. 3a6b195 : `remove unneeded files from build`
While trying to fix warnings, I discovered the file is not actually used. I cleanly removed it from build.
6. 541e298 : `[-Wclazy-detaching-temporary]`
Rationale is `constfirst` returns a read only reference instead of a reference to first element. This is important because to return a non-const reference you actually copy the whole list in memory because of how QT containers work.
7. af9f35e : `[-Wclazy-qstring-insensitive-allocation]`
Calling only one function instead of 2 gives better performances. I agree the gain is limited but it's also very readable code.
8. 3cba556 : `rename signal to avoid [-Wclazy-overridden-signal]`
`saveGeometry` name was already used somewhere else potentially causing unexpected issues because of calling wrong signal or function. 
9. 43a78e6 : `remove a unused line to fix [-Wclazy-qdatetime-utc]`
I did not dig to much about the actual warning cause. Variable is unused so I deleted it.
10. 224731a : `remove unused function with bad signature fix [-Woverloaded-virtual]`
This one is interesting. There were declaration of `draw` function (not actually implemented) that were not matching the virtual parent class declaration as they should. If one were to implement and call those functions, they may not be called as expected depending on object type. As they were actually unused and not implemented, I removed them. 
11. 8db69ca : `remove unused variables`
Unused variables of complex types were not detected by GCC. Clang did.
12. 28a8871 : `[-Wconstant-conversion]`
char is signed but matrix was declared U8 (unsigned). So 255 was actually overflowing and re-converting. Here it worked but that was not clean.
13. eb4c370 : `[-Wunused-private-field]`
As name states. An unused field in object.
14. 7018a7a : `re-enable problem matcher as there are very few warnings left`
This is to display remaining warnings in review pane. I need to fix it as the matcher is not 100% functional (this PR is still draft)
15. will be done in separate PR : `[-Wclazy-incorrect-emit]`
`emit` is used on non-signals OR `emit` is used in constructor and might have no effect
Quite concerning IMHO.

I think all these changes have valid reason to exist. Much more than the `printf` changes to `arg`. Also in previous case I had no choice because valid warnings were flooded by printf deprecation that could not be muted. Here with clazy we wan keep some changes and selectively mute warnings which is much more appropriate.

Kindly answer for each number if you would like me to keep changes or remove from the PR.

githubdoe:
Julian,

I'm sorry but the amount of your work exceeds my ability to review it in a timely manner of even forever.  I don't feel I can justify spending time on things that might improve performance or most of the warnings that you are addressing to improve the maintainability of the code by others or fix depreciated Qt features as long as they still work.

If these were done on a smaller basis like when a bug was being addressed for a particular feature then I could manage.  However as it is now I'm overwhelmed and don't want to spend the needed time it takes to review these changes.

So somehow you have to scale back these things.  At this pace it becomes a full time day job for me and I'm not so interested in doing that.   

At least that is my current thoughts.  I will work on anything that is considered a bug by you guys however.  So perhaps put the word bug or possible bug in the request.  

Sorry.  

I am leaning toward not to change to an organization.  If when the time comes and I'm not here or if you want to continue at the current pace then I guess a clone of the project is the way forward but at that point change it's name and inform people that it is a derived work.

atsju:
Thank you for this clear feedback. Not having time is absolutely a valid justification (even if @gr5 is here to help the reviews) and I have no personal problem with that. I will close non-bug PRs for the moment then.

About the clone/derive VS organization, I absolutely not want to create a derive work now. This would make no sense when a tool is already maintained. However I don't understand why do you want us to create a fork/clone or whatever when you will stop maintenance ? Having a tool you created continue to be maintained and evolve is (from my POV) greatest thing that can happen to an open source developer and his tool. 
Also note that moving to an organization now would change nothing to development. Neither me neither @gr5 want to remove you from the work as long as you maintain it and you would keep final word on everything like now. I only propose this because for example if you can't access a computer for 3weeks or worse then @gr5 can not even merge his PRs. We would then have no other choice than create a derivative work which makes me sad.

PR #76, opened 2023-08-16
[bug] fix some memory leaks

atsju:
I run Valgrind on Linux using QTcreator. Running the tool is really straightforward but incredibly slow. And you need to do every actions you would do as end user so that the actual leak is shown. This is not static analysis. And after analysis you need to actually fix the code and test again.  
There were 2 actual/growing leaks that are fixed. The rest is cleanup to reduce tool noise because too much warnings is always masking real problems. Maybe they did actual leaks.  
One of the main problems is that [object tree](https://doc.qt.io/qt-6/objecttrees.html) is not built correctly so the objects don't know when to be deleted. Of course if they exist only once, leak will not grow but this still makes noise during analysis so it must be implemented.

About contourplot.cpp, I only commented out. I need your feedback on this as I'm not experienced enough in QT and QWT.

About singleton, the C++11 way is actually cleaner and will call destructor at end of program. This helps removing noise. I recommend to update all singleton instances in the tool (I counted 15 singleton objects with `get_Instance`, `getInstance` and `get_instance` ...)
Please read https://leimao.github.io/blog/CPP-Singleton/

This would also have saved some leaks if it existed before. Enjoy the reading : https://en.cppreference.com/w/cpp/memory/unique_ptr

@githubdoe To simplify review I prefer to do many small PR instead of one big. And even like that, I recommend you look at each commit individually to simplify review. 
Also please take your time for the reviews, there is no urgency.

Sorry, the explanation was already longer than I expected. Do not hesitate if you have any remark or question.

### DIFF
--- a/contourplot.cpp
+++ b/contourplot.cpp
@@ -548,13 +548,6 @@ void ContourPlot::showSpectrogram(bool on )
     d_spectrogram->setDisplayMode( QwtPlotSpectrogram::ImageMode, on );
     d_spectrogram->setDefaultContourPen(m_do_fill ? QPen(m_contourPen) : QPen(Qt::NoPen));
 
-    //TODO JST 20230815 commented this code because seems to have no purpose. Remove if this is confirmed
-    //QwtPlotShapeItem *item = new QwtPlotShapeItem( "" );
-    //item->setItemAttribute( QwtPlotItem::Legend, true );
-    //item->setLegendMode( QwtPlotShapeItem::LegendShape );
-    //item->setLegendIconSize( QSize( 20, 20 ) );
-    //item->setRenderHint( QwtPlotItem::RenderAntialiased, true );
-
     replot();
 }
 
--- a/contourplot.cpp
+++ b/contourplot.cpp
@@ -548,11 +548,12 @@ void ContourPlot::showSpectrogram(bool on )
     d_spectrogram->setDisplayMode( QwtPlotSpectrogram::ImageMode, on );
     d_spectrogram->setDefaultContourPen(m_do_fill ? QPen(m_contourPen) : QPen(Qt::NoPen));
 
-    QwtPlotShapeItem *item = new QwtPlotShapeItem( "" );
-    item->setItemAttribute( QwtPlotItem::Legend, true );
-    item->setLegendMode( QwtPlotShapeItem::LegendShape );
-    item->setLegendIconSize( QSize( 20, 20 ) );
-    item->setRenderHint( QwtPlotItem::RenderAntialiased, true );
+    //TODO JST 20230815 commented this code because seems to have no purpose. Remove if this is confirmed
+    //QwtPlotShapeItem *item = new QwtPlotShapeItem( "" );
+    //item->setItemAttribute( QwtPlotItem::Legend, true );
+    //item->setLegendMode( QwtPlotShapeItem::LegendShape );
+    //item->setLegendIconSize( QSize( 20, 20 ) );
+    //item->setRenderHint( QwtPlotItem::RenderAntialiased, true );
 
     replot();
 }
--- a/dftarea.cpp
+++ b/dftarea.cpp
@@ -110,7 +110,7 @@ DFTArea::DFTArea(QWidget *mparent, IgramArea *ip, DFTTools * tools, vortexDebug 
     m_PSIstate = 0;
     QRect rec = QGuiApplication::primaryScreen()->geometry();
 
-    m_Psidlg = new PSI_dlg;
+    m_Psidlg = new PSI_dlg(this);
     rec.setLeft(rec.width()/6);
     rec.setTop(rec.height()/4);
     rec.setWidth(rec.width()/4);
--- a/dftarea.cpp
+++ b/dftarea.cpp
@@ -627,10 +627,6 @@ cv::Mat DFTArea::vortex(QImage &img, double low)
     double *qmap = new double[size];
     double *orient = new double[size];
 
-
-    double *imRe = new double[size];
-
-
     //double *fdom[2]; fdom[0] = new double[size]; fdom[1] = new double[size];
     double *d1[2]; d1[0] = new double[size]; d1[1] = new double[size];
     double *d2[2]; d2[0] = new double [size]; d2[1] = new double [size];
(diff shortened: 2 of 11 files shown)


githubdoe (approved):
Yes ok to delete.

githubdoe (approved):
Yes.  Code originally came from one of the other contributors and was in C.  I decided since it was so complex to just leave it without converting it into objects.  Too hard to debug if it was not correct.  

Current leak fix is ok.

githubdoe:
Not sure I have done this correctly.  I thought I was reviewing each code segment and approving only that.  So far I have only looked at 2 of the things in the list emailed to me.  If instead this is for the whole pull request then that is not approved.

punwrap.cpp
githubdoe, line 1: OK I think I have found the right buttons to press to comment on specific code.
As I said earlier this was C code that was contributed to me that I did not want to mess with too much to turn it into objects.

contourplot.cpp
githubdoe, line 1: Yes OK to delete.

atsju:
Wow it must be late for you in the US.

Yes approval is global to PR even when you look at individual commits.

For comments, you are currently commenting global files. You can also hover over the line of code you want to comment and click on the "+" it's even more accurate. 
image

contourplot.cpp
atsju: I deleted it 👍

githubdoe (commented):
The big one was I think the punwrap one.  I think that is what I have been looking for for years.  It is what caused
 the 32 bit version to run out of memory when trying to analyzed too many igrams.  
So far I think the others don't matter much.  Some happen when the app is closed and some dialogs don't get properly deleted.   Those don't eat up the memory per wavefront like the punwrap  leak did.
I will work through them all in time.  Thanks for finding the punwrap one.

atsju:
> The big one was I think the punwrap one. I think that is what I have been looking for for years. It is what caused
> the 32 bit version to run out of memory when trying to analyzed too many igrams.
> So far I think the others don't matter much. Some happen when the app is closed and some dialogs don't get properly deleted. Those don't eat up the memory per wavefront like the punwrap leak did.
> I will work through them all in time. Thanks for finding the punwrap one.

Yes punwrap was the big one. I could not find it without removing the other "1 time only" leaks because of noise they generate.
I think there are still some other, those are just the basic ones that every user would experience.

igramarea.cpp
```diff
@@ -109,7 +109,7 @@ IgramArea::IgramArea(QWidget *parent, void *mw)
     opacity = set.value("igramLineOpacity", 65.).toDouble();
     lineStyle = set.value("igramLineStyle", 1).toInt();
     setMouseTracking(true);
-    m_dftThumb = new dftThumb();
+    m_dftThumb = new dftThumb(this);
```
gr5, line 112: Why are you passing the parent pointer to this and several other widgets and dialogs?

QDialog doc seems to indicate you only need this to center the window over the window below it but the one case I looked at already it doesn't matter as the QDialog is embedded in a larger window.

QWidget says if parent is null it will be a window and if it has a parent will graphically constrain things within the parent windows border.

I guess I don't understand the implications.

atsju: TBH I did not see a visual difference at execution but I'm not 100% sure of my understanding.

Main implication that made me change this code is : https://doc.qt.io/qt-6/objecttrees.html  
TLDR, a QObject without parent cannot be deleted by parent death.

Now that I re-read the documentation, I understand that dftThumb being a QDialog, this means a popup so I'm not sure we care about having parent for memory leak. So parent would only place the widget at center instead of random place ...  
About Qdialog : 

> Note that QDialog (and any other widget that has type Qt::Dialog) uses the parent widget slightly differently from other classes in Qt. A dialog is always a top-level widget, but if it has a parent, its default location is centered on top of the parent's top-level widget (if it is not top-level itself). It will also share the parent's taskbar entry.

I will need to test this again with Valgring when I'm home. For Qwidget it's propably required to do what I did. And for Qdialog not required but small difference. As I said, I will try. Thank you for heads up.

atsju: OK I read the code and I think I understand now. 
Indeed all those I modified are Qdialog, so adding a parent should not be necessary. However they were all using a format like `m_Psidlg = new PSI_dlg;`. And the destructor does not `delete m_Psidlg;` and this is what generated noise (1 time leak).  
I think both solutions are good. Either `m_Psidlg = new PSI_dlg(this);` or `delete m_Psidlg;`. The later might be more complex and need to check if object is already dead before delete, I don't really know.

Is this precise enough to cancel your doubt or shall I go more in depth ? (Because I agree I do not have all 100% of the understanding right now)

gr5: It's fine for the QDialogs.  I'm concerned about QWidgets.  I thought there was at least one... let me find it...

Okay, my mistake.  I think they are all QDialogs.  I'll double check them all.  

So it's fine as is.

gr5 (approved):

gr5:
Dale, I looked through and approve the changes - are you done reviewing?  I'm ready to create a new tag and a new version and do some testing before releasing a new version.  Some of the changes (I'll write them all up):
- created a new oln file format (and hopefully reduce crashes)
- fixed a memory leak

githubdoe:
No I am not done.  It will be some time I expect.  I will comment when done.

dftarea.cpp
```diff
@@ -639,18 +635,18 @@ cv::Mat DFTArea::vortex(QImage &img, double low)
...
-
-
-
     if (m_vortexDebugTool->m_showInput){
-        cv::Mat tmp = cv::Mat(ysize,xsize,numType, imRe);
-        cv::Mat xx;
-        tmp.convertTo(xx,CV_32F);
-        cv::imshow("input", xx);
-        cv::waitKey(1);
+        double *imReTmp = new double[size];
```
githubdoe, line 639: Ok but only used by the developer contributor to help him debug the vortex algorithm and then by me for understanding of how it worked.  So never really gets called except by curious users.

githubdoe (approved):
Completed theses commit reviews I think.  All is good.

dftarea.cpp
atsju: No, you are wrong here. I understand the intent but this was actually a memory leak before. And a quite big one (I remember this was not a one time leak in Valgrind).
The `imre=new` was done outside the `if(debug)` and then there was a later `imre=new` thus leaking by losing the address of first allocated block without freeing it before.

To fix I made the choice to create this `imReTmp` variable instead of using `imRe` for 2 different usages. And to acllocate only if actually needed.

dftarea.cpp
```diff
@@ -738,7 +734,7 @@ cv::Mat DFTArea::vortex(QImage &img, double low)
 
     imPlanes[1] *= 0.;
 
-    imRe = (double *)(imPlanes[0].data);
+    double *imRe = (double *)(imPlanes[0].data);
```
atsju, line 737: @githubdoe  the leak actually happened here by overwriting original `imRe`

atsju:
thanks you both for you reviews !

PR #78, opened 2023-08-20
[bug] fix change in window behavior introduced in 90f7bbd

atsju:
- `igramIntensity` `ColorChannelDisplay` `PSI_dlg` I removed the this as they are actual dialog my understanding is that they should either be modal and have a parent or not be modal and not have parent 

Documentation states 
> The parent relationship of the dialog does not imply that the dialog will always be stacked on top of the parent window

But I searched a lot and I did not find how to use parent, be non-modal and not always stay on top alltogether
It's probably better as it is now. And I agree with Dale, `igramIntensity` `ColorChannelDisplay` windows should not be modal. They are updated on the fly during process and modal would lock out user.
For `PSI_dlg` however, maybe it should be full modal. I let you think about it as I have no experience with the PSI_dlg usage.

Also about calling the destructor, it does not only free memory properly but also makes sure the windows are closed when mainWindow is closed. Dialog staying open was a bug already present in 6.2 and 6.3.1. Thanks Georges for pointing it out. Maybe add it to release note.

- `dftthumb` I only changed the display as it should probably have a parent, not like the other windows

### DIFF
--- a/dftarea.cpp
+++ b/dftarea.cpp
@@ -110,7 +110,7 @@ DFTArea::DFTArea(QWidget *mparent, IgramArea *ip, DFTTools * tools, vortexDebug 
     m_PSIstate = 0;
     QRect rec = QGuiApplication::primaryScreen()->geometry();
 
-    m_Psidlg = new PSI_dlg(this);
+    m_Psidlg = new PSI_dlg(nullptr);
     rec.setLeft(rec.width()/6);
     rec.setTop(rec.height()/4);
     rec.setWidth(rec.width()/4);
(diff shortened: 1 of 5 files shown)


githubdoe (approved):
PSIDialog should be modeless IIRC. It is both a help window and provides some controls for the complex PSI analysis process.  PSI is used by about 2 people.  Maybe

gr5 (approved):

gr5:
I'm going to test this and if it passes all my tests I will want to release this version as 7.0.0.

The last time I thought I could test in just 30 minutes.  I was very mistaken.  There's really a lot to test.  :)

PR #83, opened 2023-08-23
update release note for release 7.0.0

atsju:
What do you think of reversing the order of file ? Most interesting part is at end of file right now and still growing.

like start with 7.0.0
and then go down 6.3.1
...
1.0

### DIFF
--- a/RevisionHistory.html
+++ b/RevisionHistory.html
@@ -342,4 +342,29 @@ Displays error of pixel under contour plot cursor and move the profie angle to m
 <li>DFTF can now build and run on Linux (build is only tested on latest Ubuntu)</li>
 <li>Move to semver format for version names</li>
 </ul>
+</ul>
 
+<ul><li>Version 7.0.0</li>
+<ul>
+<li>Remove compilation warnings that only affect DFTFringe programming team</li>
+<li>Project is now buildable on macOS</li>
+<li>Fixed all bugs related to rotating and averaging when there are regions or central obstruction</li>
+<ul>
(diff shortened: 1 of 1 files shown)


RevisionHistory.html
```diff
@@ -342,4 +342,29 @@
...
+</ul>
+<li>Fixes GUI issues related to outline (oln) file</li>
+<ul>
+    <li>Oln files are now stored in a new format. DFTFringe can still open files from previous format</li>
+    <li>Fixed bug where oln files can be read wrong (typically 12% of the time)</li>
+    <li>Fixed bug where reading and writing to oln file after cropping would move outline and region locations</li>
+</ul>
+<li>Fix some memory leaks</li>
+<li>Hovering over data in "astig stats" dialog will show picture name</li>
+<li>Automatically close some secondary windows that where not closing at end of main application</li>
```
gr5, line 367: "where" should be "were"

gr5: I prefer newest at top.  Not sure if @githubdoe cares either way?  I suspect no one reads the revision history anyway so maybe leave it as is.  I'll be putting this exact same text on the release page and that's more where people will see it.

atsju: If nobody read them then we should delete the file... Would cost less maintenance ^^

atsju, line 367:
```suggestion
<li>Automatically close some secondary windows that were not closing at end of main application</li>
```

gr5 (approved):

Issue #82, opened 2023-08-23
many dialogs don't close when application closes


gr5:
**To Reproduce**
**zernike smoothing**

**For the following, first process an igram to get a wavefront.  Then:**

1. Do "tools" "smooth current wave front".  Before doing anything with this dialog close the app.  The dialog stays running and the DFTF icon is still visible as an actively running application.

1. tools  wavefront inspector
2. view show statistics of loaded wavefronts

**needs igram loaded but not a wavefront**
1.tools outline helper

**others - just launch dialog and close window:**
1. tools user color map
1. tools test stand astig removal
3. tools camera calibration
4. tools astig stats
5. config mirror
6. config preferences

tools tilt versus astig analysis - first you need to create a zernike csv file.  With one or more wavefronts loaded do "file" "save zernike values in csv".  Then run the astig analysis off the csv file.

**Expected behavior**
all dialogs should close when closing DFTF main window.

**Desktop (please complete the following information):**
win11, master branch as of 8/23/23 - probably will become v7.0.0

atsju:
This one is for me I think. I will love to write all these clean destructors :)

atsju:
### wft needed

- [ ] 1. Do "tools" "smooth current wave front". Before doing anything with this dialog close the app. The dialog stays running and the DFTF icon is still visible as an actively running application.
- [ ] 2. tools wavefront inspector
- [ ] 3. view show statistics of loaded wavefronts
- [ ] 4. Click full screen above the 3d window, alt-tab to app, close app
- [ ] 5. right click upper right wavefront coutour chart. Choose "full screen". alt-tab to app and close app.

### needs igram loaded but not a wavefront
- [x] 1. tools outline helper

### others - just launch dialog and close window:

- [x] 1. tools user color map (#84)
- [x] 2. tools test stand astig removal (#93)
- [x] 3. tools camera calibration (#95)
- [x] 4. tools astig stats => not reproduced on Linux and Windows with v7.0.0
- [x] 5. config mirror (#96)
- [x] 6. config preferences (#102)

atsju:
at e25f5405effce2c3aa6a444e45bf918151102c71 with every preidentified problems already fixed.
AI found 6 remaining user-facing cases in this branch that can still behave like orphan windows after closing the main app.

To investigate manually each of them.

# Astig stats distribution popup
Code evidence: astigstatsdlg.cpp:729, astigstatsdlg.cpp:934, opened from mainwindow.cpp:1549
[...]

PR #84, opened 2023-08-26
update SingleApplication, move sources and icons around, deletes unused icons and files

atsju:
- I started working on #82 so it will fix userColorMapDlg closing on app closing.
- I had this one time memory leak from SingleApplication and I noticed this was [well maintained and this had been fixed](https://github.com/itay-grudev/SingleApplication/blob/master/CHANGELOG.md) so I updated to latest release.
- I separated external sources (bezier and singleApplication) to their own folder with their associated licence file.
Sources display well in QTcreator
image
- I deleted unused icons (their are in git history if we need one back some day)
- I moved all icons from base folder to /res folder
- Installer does not distribute icons anymore (only the ones used in HTML help files)
- removed mikeSpecsInterface from build (but kept file)
- deleted spline.h (confirmed by Dale)


atsju:
I got back to one line per source in the .pro
Much easier to review and less merge conflicts

gr5 (approved):

gr5:
When you move or rename a file, usually git figures it out as the old and new file are identical.  But if you also make an edit or maybe for reasons I don't understand, sometimes git marks the old file as deleted and the new file as created.

So in the future, please use git mv  (https://git-scm.com/docs/git-mv)

This way git knows you didn't intend to just delete and create but that you are renaming or moving a file.  This will help us follow edit history in git better (git will be smarter about it and show you historical edits made when the file had the old name or location), and also make reviews simpler.

I mean all the edits are still there but harder to find because the file that was edited was "deleted" (sometimes, if you don't use git mv).

gr5:
> * I started working on [many dialogs don't close when application closes #82](https://github.com/githubdoe/DFTFringe/issues/82) so it will fix userColorMapDlg closing on app closing.

Wow you got distracted?  You only fixed one of them?  lol.  Okay.

atsju:
> So in the future, please use git mv (https://git-scm.com/docs/git-mv)
I suppose you are talking about single_application. I had a look commit by commit and it has been modified in one and moved in an other. `git mv` would have changed nothing I think.

> Wow you got distracted? You only fixed one of them? lol. Okay.
Yes

gr5:
> I suppose you are talking about single_application. I had a look commit by commit and it has been modified in one and moved in an other. `git mv` would have changed nothing I think.

Yes.  Well I guess it's just a github thing then.  Yeah I see it now in my git tools that it considered it a rename in one commit.  But I guess github didn't know how to show it as a rename *and* an edit.

PR #86, opened 2023-09-01
[bug fix] "show only average" in astig stats wasn't removing display

### LINKED ISSUE #85: broken "show only average" in astig stats
**Describe the bug**
"show only average" does still show all points.
image
probably introduced by my change in #75 :( 

**Desktop (please complete the following information):**
 - OS: Win
 - DFTFringe version: v7.0.0
 - Build: official

atsju:
I'm getting better at QWT and QT. Thank chatGPT on this one also.

fix #85 by using QWTplotPicker to display what we need to. I have been able to fix the bug and still keep the feature.
Moving mouse +/-5 pixels around the data will display the wavefront name

### DIFF
--- a/astigstatsdlg.cpp
+++ b/astigstatsdlg.cpp
@@ -173,8 +173,6 @@ astigStatsDlg::astigStatsDlg(QVector<wavefront *> wavefronts, QWidget *parent) :
 
 astigStatsDlg::~astigStatsDlg()
 {
-    delete zoomer; //TODO JST 2023/08/30 check if delete is needed. I don't think so because the parent is set
-
     delete ui;
 }
 class measure{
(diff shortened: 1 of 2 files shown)


gr5 (approved):
Code looks fine.  I tested it also.  Works great.

Should I wait to m;erge this until after I'm done with spdlog?  Might be a few more days.

atsju:
> Should I wait to merge this until after I'm done with spdlog? Might be a few more days.

This one is unrelated. You can merge whenever you want.

The other 2 might create a small merge conflict because the .pro oneliner source list introduced in MacOS support is not a so good idea in the end. But you can still merge in any order and I will fix it.

PR #88, opened 2023-09-09
[feature] Enter displays wavefront. Double click and F2 rename from the start

### LINKED ISSUE #87: [BUG] wavefrontList is bugged and could be improved
**Describe the bug**
Double click on a wavefront does no permit to rename it.
Once you renamed one wvft with right click they become editable/renamable with double click.

**Improvement**
I would like to be able to rename using F2 which is standard renaming shortcut on windows
Up/down + enter should select/display the wavefront. Today a click is mandatory and this bothers me as a user.

I'm already working on the code so this issue is more about getting a feedback while working on it.

**My basic proposal (this is all nativelly supported by QListWidget)**
- Add F2 to rename
- Add up/down + enter as a way to display a wavefront

**Potentially controversial proposal**
- Right click is for context menu. I propose that we remove the right click as renaming. Replaced by both F2 and double click. It's more natural even if you used right click for years.
- I have seen in the code that simple click and double click on waveform have different effect. I did not dig enough to make clear recommendation about it right now but I think all should be managed by simple click and remove double click effect.

Both the double click and right click are a bit of masked features that a user will not naturally find if he doesn't know about it.

atsju:
fix #87 

- [x] press enter displays wft
- [x] F2 renames wft
- [x] ~~right click does not rename anymore~~ Do a real right click menu instead as per George recommandation
- [x] double click has no special effect anymore
- [x] greying out wft names is strange behavior
- [x] create a shortcut for delete wft (dell key)
- [x] any edit will trigger correct signal to reload the wft and have name displayed correctly everywhere
- [x] right click menu can rotate and invert
- [x] fix display bug on delete

I think I achieved something very nice.
Now the actually displayed wft has a small icon in front. I think this is what Dale wanted to achieve with gray/black foreground but it was not 100% working and not always displaying what it should on multi-selection.

Also fixed a bug where wft name could be wrong on metrics window
image

image

### DIFF
--- a/surfaceanalysistools.cpp
+++ b/surfaceanalysistools.cpp
@@ -56,8 +56,8 @@ surfaceAnalysisTools::surfaceAnalysisTools(QWidget *parent) :
     connect(ui->wavefrontList->itemDelegate(), SIGNAL(closeEditor(QWidget*, QAbstractItemDelegate::EndEditHint)), this,
             SLOT(ListWidgetEditEnd(QWidget*, QAbstractItemDelegate::EndEditHint)));
     ui->wavefrontList->setContextMenuPolicy(Qt::CustomContextMenu);
-    QShortcut* dellShortcut = new QShortcut(Qt::Key_Delete, this);
-    connect(dellShortcut, &QShortcut::activated, this, &surfaceAnalysisTools::on_deleteWave_clicked);
+    QShortcut* delShortcut = new QShortcut(Qt::Key_Delete, this);
+    connect(delShortcut, &QShortcut::activated, this, &surfaceAnalysisTools::on_deleteWave_clicked);
 }
 
 void surfaceAnalysisTools::enableControls(bool flag){
@@ -107,7 +107,7 @@ void surfaceAnalysisTools::on_spinBox_2_valueChanged(int arg1)
     emit centerMaskValue(arg1);
 }
 
-// this is responsible of givin user feedback about which wft is currently displayed
+// this is responsible of giving user feedback about which wft is currently displayed
 void surfaceAnalysisTools::currentNdxChanged(int ndx){
     if (ui->wavefrontList->count() == 0)
         return;
--- a/surfaceanalysistools.cpp
+++ b/surfaceanalysistools.cpp
@@ -17,7 +17,9 @@
 ****************************************************************************/
 #include "surfaceanalysistools.h"
 #include "ui_surfaceanalysistools.h"
+#include <QAction>
 #include <QDebug>
+#include <QMenu>
 #include <QSettings>
 #include <QtAlgorithms>
 #include <QLineEdit>
@@ -73,9 +75,8 @@ void surfaceAnalysisTools::addWaveFront(const QString &name){
         shorter = list[list.size()-2] + "/" + list[list.size()-1];//fontMetrics().elidedText(name,Qt::ElideLeft,w);
 
     ui->wavefrontList->addItem(shorter);
-    lastCurrentItem = ui->wavefrontList->count()-1;
+    currentNdxChanged(ui->wavefrontList->count()-1);
     ui->wavefrontList->item(lastCurrentItem)->setFlags(Qt::ItemIsEditable | Qt::ItemIsEnabled | Qt::ItemIsSelectable);
-
 }
 
 void surfaceAnalysisTools::removeWaveFront(const QString &){
@@ -93,12 +94,6 @@ void surfaceAnalysisTools::on_blurCB_clicked(bool checked)
     emit surfaceSmoothGBEnabled(checked);
 }
 
-void surfaceAnalysisTools::on_wavefrontList_itemDoubleClicked(QListWidgetItem *item)
-{
-    ui->wavefrontList->editItem(item);
-    emit wavefrontDClicked(item->text());
-}
-
 void surfaceAnalysisTools::on_spinBox_valueChanged(int arg1)
 {
     emit outsideMaskValue(arg1);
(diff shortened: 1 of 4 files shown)


gr5:
I love this but think you should keep the right click functionality for people who expect that.  "delete wft" shortcut should be "Delete" key, right?  Unless you plan to pop up a context menu on right click with "rename" as one of the options.

I am the shortcut key master as I hate using the mouse.  I know probably 1000 shortcuts in windows.  But I didn't know about F2 for rename until I started using Ubuntu as my primary computer.  I was shocked that it works in windows as well.  Now it's the only shortcut I use for renaming because it works in both Windows and Linux.

atsju:
I think I'm done with this one.
Lots of small improvement for end user. Little changes for users used to their interface. Probably fixed one or two display bugs in the process.
I refrained from doing any architectural change.

surfaceanalysistools.cpp
```diff
@@ -246,24 +288,23 @@ void surfaceAnalysisTools::on_InvertPB_pressed()
...
-    ui->wavefrontList->item(t.row())->setSelected(true);	// even a right click will select the item
-    ui->wavefrontList->editItem(item);
-}
-
-void surfaceAnalysisTools::ListWidgetEditEnd(QWidget *editor, QAbstractItemDelegate::EndEditHint ){
-    QString NewValue = reinterpret_cast<QLineEdit*>(editor)->text();
-    QModelIndexList indexes = ui->wavefrontList->selectionModel()->selectedIndexes();
-    if (indexes.length() == 1){
-        emit wftNameChanged(indexes[0].row(), NewValue);
+// this is troggered on renaming/editing end
```
gr5, line 291: troggered -> triggered

surfaceanalysistools.cpp
```diff
@@ -110,46 +107,91 @@ void surfaceAnalysisTools::on_spinBox_2_valueChanged(int arg1)
...
+
+    // Create "Invert" action
+    QAction *invertAction = new QAction("Invert", this);
+    connect(invertAction, &QAction::triggered, [this, item]() {
+        on_InvertPB_pressed();
+    });
+    contextMenu.addAction(invertAction);
+
+    // Create "Rename" action
+    QAction *renameAction = new QAction("Rename", this);
```
gr5, line 178: It would be nice if this also showed the shortcut key as F2.  But not mandatory.  Similar for any other actions (delete?) that have a shortcut key.

atsju: Like this ? :

```
Delete (Dell)
Invert
Rename (F2)
Rotate
```

surfaceanalysistools.cpp
atsju:
```suggestion
// this is triggered on renaming/editing end
```

surfaceanalysistools.cpp
gr5: Screenshot from 2023-09-12 16-25-24

"Del" should have one "l" and there must be a way to right justify.  This is how it's done in linux (ubuntu) and alsow windows.  And also Mac.  Right justify the shortcuts.

gr5:
Found a bug.  I created a wavefront.  I rotated it.  Second I renamed to ccw.  I then averaged.  Now I have 3 wavefronts.

Now I **deleted** the last one but it's still there I think?  Anyway I did an average of the first two and now I get a blank wft in between where the old "average" was.  When I click on the blank wavefront it is indeed an average just like the one I had deleted.

Capture

gr5:
I really love the green circle.  

There is a feature where Dale does averaging and if there is a central obstruction, it itakes the obstruction from the first wavefront (I think the one with the green dot as I think it's really the last one clicked maybe if I remember right).

Also the wft that is "first" - the resolution of that wavefront is used for the final average.  Usually the resolution is the same but when you are averaging wavefronts from different people of the same mirror (like the round robin) then it's helpful to be able to pick the resolution.

Anyway, for that reason and many others, I love the green circle.

atsju:
> Found a bug. I created a wavefront. I rotated it. Second I renamed to ccw. I then averaged. Now I have 3 wavefronts.
> 
> Now I **deleted** the last one but it's still there I think? Anyway I did an average of the first two and now I get a blank wft in between where the old "average" was. When I click on the blank wavefront it is indeed an average just like the one I had deleted.
> 
> Capture

Oh nice catch. Please reproduce and tell me exactly where to click or select. Some of those bugs are really sensitive to what was selected, clicked,... 
I will go to bed now but will have much time to fix this WE. I will be happy if you find other corner cases. 
Also try multi-selections and selections with arrow keys. 4 eyes will catch more than me alone

gr5:
I'm having trouble duplicating!  Still trying.  But I got distracted because "the bug" happened and FINALLY it happened with your new logging and FINALLY I have a massive clue as to the problem.  Massive.

gr5:
It's not a bug!  I got it to duplicate but it's user error I think.  I think I selected rename but then changed my name and hit "del" to test out the delete shortcut and ended up giving the wavefront a blank name.

surfaceanalysistools.cpp
gr5: Okay so let's at least do your suggestion with the parenthesis please "()".  And I'll approve.  Or if you can do "right justfied" text than even better!

surfaceanalysistools.cpp
```diff
@@ -53,6 +56,8 @@ surfaceAnalysisTools::surfaceAnalysisTools(QWidget *parent) :
     connect(ui->wavefrontList->itemDelegate(), SIGNAL(closeEditor(QWidget*, QAbstractItemDelegate::EndEditHint)), this,
             SLOT(ListWidgetEditEnd(QWidget*, QAbstractItemDelegate::EndEditHint)));
     ui->wavefrontList->setContextMenuPolicy(Qt::CustomContextMenu);
+    QShortcut* dellShortcut = new QShortcut(Qt::Key_Delete, this);
```
atsju, line 59:
```suggestion
    QShortcut* delShortcut = new QShortcut(Qt::Key_Delete, this);
```

surfaceanalysistools.cpp
```diff
@@ -110,46 +107,91 @@ void surfaceAnalysisTools::on_spinBox_2_valueChanged(int arg1)
     emit centerMaskValue(arg1);
 }
 
- void surfaceAnalysisTools::currentNdxChanged(int ndx){
-     if (ui->wavefrontList->count() == 0)
-         return;
+// this is responsible of givin user feedback about which wft is currently displayed
```
atsju, line 110:
```suggestion
// this is responsible of giving user feedback about which wft is currently displayed
```

surfaceanalysistools.cpp
```diff
@@ -110,46 +107,93 @@ void surfaceAnalysisTools::on_spinBox_2_valueChanged(int arg1)
...
+        on_InvertPB_pressed();
+    });
+    contextMenu.addAction(invertAction);
+
+    // Create "Rename" action
+    QAction *renameAction = new QAction("Rename", this);
+    connect(renameAction, &QAction::triggered, [this, item]() {
+        ui->wavefrontList->editItem(item);
+    });
+    renameAction->setShortcut(QKeySequence(Qt::Key_F2));
```
atsju, line 183: @Interferometry could you please test this branch on macOS ?
And let me know if you can rename the wft with F2 or only enter ? Then I can put an ifdef for macOS.
Unfortunately there is no QKeySequence::Edit so I need to put the correct key for each platform manually.
https://codebrowser.dev/qt6/qtbase/src/widgets/itemviews/qabstractitemview.cpp.html#2497
@gr5 it can still be merged as is. We can fix this later if needed.

surfaceanalysistools.cpp
gr5: I love it now!  This will help new users quite a bit.

gr5:
I have to admit, the moving green dot confuses me briefly every time because everything is bouncing around left and right.  I mean it's obvious what's going on when you think about it but while it bounces around it seems almost like they are swapping vertically.

I thought I would get used to it but I've been playing with this quite a bit now.

I love the function of the "Enter" key now as well!  I can move around with arrows and hit enter to load that wavefront.  It's really nice.

What do you think of putting a blank space in front of the rows without the green dot?  So they don't shift left/right?

atsju:
Might not be cleanest way to do it as it uses a fixed size but it works

gr5:
Using an empty icon to align with the green-circle icon works fine for me.

gr5 (approved):

PR #89, opened 2023-09-09
update armadillo

atsju:
should fix warnings ?


atsju:
sorry this was supposed to be openned on my side locally.

It's not needed. And warnings are due to old compiler, not to armadillo itself

Issue #91, opened 2023-09-14
Sometimes regions fill in with data off by many wavelengths.


gr5:
**To Reproduce**
Create a region somewhere in the igram.  Process normally.  Repeat hundreds of times restarting DFTF each time until you get a massive multi-wavelength spike in the region.  The whole region.

**Expected behavior**
A smooth area where the region outline was.

**Screenshot of the issue**
Notice how my region (squarish spot) is many (48.1357 to be exact when looking at debug output) waves too high in this example.  Strange that term 48 was a value close to 48 but another time it was 111.943 so just a coincidence I guess.
![thebug](https://github.com/githubdoe/DFTFringe/assets/5414808/df60cf89-9fac-4673-bde3-6a47de21e0db)

So I figured out the issue.  The fillVoid() is using 49 zernike terms but the last one is sometimes a crazy value.  The fix will be to limit to Z_TERMS zernike values when filling in the voids.  I don't know why the last one is a crazy value but I'm just going to limit smoothing to Z_TERMS.

I call it "smoothing" because it's almost the same code as the zernike smoothing code except it only uses the existing, available zernike values.

This took many hundreds of tries before I was lucky enough to get the bug to happen (was testing something unrelated) and *also* we had logging running.  I wouldn't have found this without Julien's new logging feature.

gr5:
Fixed in PR #92 

closing

Issue #48, opened 2023-07-19
Possible bug


gr5:
**Describe the bug**
This line of code in zernikeprocess.cpp is probably wrong:
`            if (mask.at<bool>(x,y))`
All the other times the at() function is used it's y,x.  not x,y.  This probably only affects regions as outlines are symmetrical across the y=x symmetrical axis.  Regions are not.  So no one would notice the bug unless they use regions feature.

I'm not certain but I think this code is used in the zernike smoothing feature which is relatively new and probably not used by very many people.

atsju:
permalink : https://github.com/githubdoe/DFTFringe/blob/1daf63a5dba50d9512b41a11263c122a84df85af/zernikeprocess.cpp#L1084

gr5:
Dale can you click on the link above and confirm that it should by "y,x" and not "x,y"?  This appears to be in your "new" zernike smoothing feature.  Maybe it doesn't matter?  Because the mask is circular and perfectly centered in the matrix?  Or maybe this messes up some pixels on the edge of the outline?

githubdoe:
That sounds like a reasonable assumption.  CV:Mat are accessed by row order (y) first.  By habit I tend to write (x,y) however.

gr5:
Should I just swap it then?  I can try to test but I'm not sure how to prove it's currently wrong.  After swapping I would of course test the feature and I can add a large region but I don't think I'll see any difference as the "region" is 47 zernike smoothed and the rest of the wavefront would be something like 300 zernike smoothed.

Also I think the region smoothing removes the regions from the wavefront after the smoothing is complete?  I'm not sure of that detail.  And if that's true then regions won't affect this bug and only the outline.

I vote to just swap it, do some limited testing.  What do you think Dale?

githubdoe:
I think it is a bug.  However it does not matter since the mask is symmetric.   So I think you will not see any difference unless you can trick it with a region but like you say.

PR #92, opened 2023-09-14
Fixes issue #91.  This bug was bothering me for a few months.  

gr5:
The final zernike term (49th term aka #48 aka "5th spherical") was sometimes a huge value.

I don't know if I'm hiding some other bug where the last entry in the zernike terms array gets overwritten or if the last value is never calculated in the first place.  But if it's some other bug, then Dale seems to have also hidden the bug by using 48 terms instead of 49 terms in the Zernike table on the left side of DFTF.

Z_TERMS is 48 and not 49 (49 would be 7*7 which would be more typical qty of terms).  Z_TERMS seems to limit how many
terms are displayed in the left side of the gui.

Anyway, this commit fixes the problem and will work just fine even if M_norms.size() changes some day.

Please see issue #91 for screenshot and more details.

I removed the logging as it is no longer needed now that I understand the bug.

### DIFF
--- a/zernikedlg.h
+++ b/zernikedlg.h
@@ -27,7 +27,7 @@
 #include <vector>
 #include <QVector>
 
-#define Z_TERMS 48
+#define Z_TERMS 49
 
 namespace Ui {
 class zernikeDlg;
--- a/zernikeprocess.cpp
+++ b/zernikeprocess.cpp
@@ -858,7 +858,7 @@ void zernikeProcess::fillVoid(wavefront &wf){
                         zpolar.init(rho,theta);
                         double v = 0.;
 
-                        for (int z = 0; z < wf.InputZerns.size(); ++z){
+                        for (size_t z = 0; z < wf.InputZerns.size(); ++z){
                             v += wf.InputZerns[z] * zpolar.zernike(z,rho, theta);
                         }
                         wf.data.at<double>(y,x) = v;
@@ -910,7 +910,7 @@ void zernikeProcess::fillVoid(wavefront &wf){
                     zpolar.init(rho,theta);
                     double v = 0.;
 
-                    for (int z = 0; z < wf.InputZerns.size(); ++z){
+                    for (size_t z = 0; z < wf.InputZerns.size(); ++z){
                         v += wf.InputZerns[z] * zpolar.zernike(z,rho, theta);
                     }
                     wf.data.at<double>(y,x) = v;
--- a/zernikeprocess.cpp
+++ b/zernikeprocess.cpp
@@ -826,10 +826,6 @@ void zernikeProcess::fillVoid(wavefront &wf){
         int endx = x;
         int starty = y;
         int endy = y;
-        qDebug() << "regions to fill: " << wf.regions.size();
-        for (int z = 0; z < m_norms.size(); ++z){
-            qDebug() << "zern "<< z << " " << wf.InputZerns[z];
-        }
         for (int n = 0; n < wf.regions.size(); ++n){
 
             for (std::size_t i = 0; i < wf.regions[n].size(); ++i){
@@ -862,6 +858,7 @@ void zernikeProcess::fillVoid(wavefront &wf){
                         double v = 0.;
 
                         for (int z = 0; z < m_norms.size(); ++z){
+                            if (z >= Z_TERMS) break; // fixes issue #91
                             v += wf.InputZerns[z] * zpolar.zernike(z,rho, theta);
                         }
                         wf.data.at<double>(y,x) = v;
@@ -914,6 +911,7 @@ void zernikeProcess::fillVoid(wavefront &wf){
                     double v = 0.;
 
                     for (int z = 0; z < m_norms.size(); ++z){
+                        if (z >= Z_TERMS) break;
                         v += wf.InputZerns[z] * zpolar.zernike(z,rho, theta);
                     }
                     wf.data.at<double>(y,x) = v;
--- a/zernikeprocess.cpp
+++ b/zernikeprocess.cpp
@@ -826,6 +826,7 @@ void zernikeProcess::fillVoid(wavefront &wf){
         int endx = x;
         int starty = y;
         int endy = y;
+
         for (int n = 0; n < wf.regions.size(); ++n){
 
             for (std::size_t i = 0; i < wf.regions[n].size(); ++i){
@@ -857,8 +858,7 @@ void zernikeProcess::fillVoid(wavefront &wf){
                         zpolar.init(rho,theta);
                         double v = 0.;
 
-                        for (int z = 0; z < m_norms.size(); ++z){
-                            if (z >= Z_TERMS) break; // fixes issue #91
+                        for (int z = 0; z < wf.InputZerns.size(); ++z){
                             v += wf.InputZerns[z] * zpolar.zernike(z,rho, theta);
                         }
                         wf.data.at<double>(y,x) = v;
@@ -910,8 +910,7 @@ void zernikeProcess::fillVoid(wavefront &wf){
                     zpolar.init(rho,theta);
                     double v = 0.;
 
-                    for (int z = 0; z < m_norms.size(); ++z){
-                        if (z >= Z_TERMS) break;
+                    for (int z = 0; z < wf.InputZerns.size(); ++z){
                         v += wf.InputZerns[z] * zpolar.zernike(z,rho, theta);
                     }
                     wf.data.at<double>(y,x) = v;


githubdoe (commented):
Well we have to think a little deeper.  Why does the size of m_norms get bigger than  Z_TERMS.
That happens when someone uses zernike smoothing.  So how does that work with the fill void function.  Does that function ever get called during zernike smoothing?

Zernike smoothing of course creates a wave front  that then is sampled to create the usual Zernike values that are displayed.    Think about that a bit and see if there are any problems associated with that fix.  Perhaps the m_norms need to be resized back down to the correct length after smoothing is performed.  I don't know.  But leave it to you to figure out.

Concerning the 49 or 48th term displayed or not.  I could be off by one in either the display or the computations.  There was a time when I added extra spherical terms thinking that would be a good thing.  However Mike Peck told me that was a bad idea to include just extra spherical terms without adding the associated terms in between.  So I removed them a long time ago but no longer know if it is ok to end on a spherical  term or if it is the next term in the start of the next order of terms.   I think you need to figure that out as well.  It could be in the middle of an order of terms.  I just don't remember.

gr5:
Okay I'll look at it some more.  It's hard to find where the terms are limited as everything points to 49 terms but it only displays 48 in the gui.  I don't think I used the "smoothing" function but it's possible I started running that feature but didn't finish using it because I was testing other stuff.  I just don't remember.  So I'll also do a little more testing to see if the problem happens every time I do smoothing but I doubt that is the case.

I probably didn't touch the zernike smoothing feature at all.

There is some math that takes the order (12) and then divides by 2 (6) and then adds one (7) and then squares it to get 49 zernike terms.  I don't know how it then at some point gets reduced to 48.

gr5:
Okay so the code was definitely going beyond the end of the array:

```
    for (int z = 0; z < m_norms.size(); ++z){
        v += wf.InputZerns[z] * zpolar.zernike(z,rho, theta);
    }
```
In the original code above, EVERY time I put a breakpoint on those lines (regardless of if I do zernike smoothing which doesn't run the code above before I process another igram),  m_norms.size() is 49 and InputZerns.size is 48.

So the loop above always goes past the end of the InputZerns array.

I could fix this by replacing m_norms.size() with wf.InputZerns.size() and that's what I plan to do but I will try to figure out how InputZerns gets shrunk by one.  I haven't found that line of code but still looking.

gr5:
Okay I checked in a better fix.  If you look at "files changed" tab above you will see my fix is better now.

I also understand the bug much better - **the for loop was using the wrong array (sized 49) to index through a smaller array (sized 48).**  Well technically it is a "vector", not an array.

49 is a more appropriate size for the zernike terms but 48 is of course fine as well.  Usually the size should be a perfect square (49) but that's not really important.

Here is the line of code that sets the zernike terms to 48 instead of 49 (which is the size of the m_norms array):
https://github.com/githubdoe/DFTFringe/blob/c5114acbbc5f5f77bb62b237920dd96fa39e1e44/zernikeprocess.h#L107

Z_TERMS is hard coded to 48.  This function is called without a second parameter (always) from surfacemanager.cpp.  unwrap_to_zernikes uses the second parameter to set the size of the wf.InputZerns vector (along with many other matrixes).

So I think this bug fix is better written now and I understand it better.

githubdoe:
Your comment about Z_TERMS being 49 or 48 concerns me.  It may indeed be off by one.    It is the old problem of counting vector elements in C because the first element is index 0 and not 1.

So is Z_TERMS the index of the last element or the number of elements?  I don't remember.  You need to look to see how it is used to size the inputzern arrays in the first place I think.

gr5:
I'm quite aware of the fact that arrays/vectors start at element 0 and so if it were sized to hold 49 numbers the final value would be index 48.  But that's not the issue.

Z_TERMS is the size of the array/vector/matrix.  it's 48.

You even have a translation function that translates the 49th term (index 48) to "5th Spherical" but it stays invisible in the GUI because Z_TERMS is 48 and not 49 (I didn't check for sure that it is as simple as changing Z_TERMS but that would certainly be a required step).

Zernike terms of order 12 should be 49 zernike terms.  not 48.  But I truly believe that no one uses Zernike terms after the 4th spherical and so having 36 terms (order 10) would be more than enough.

I know your Zernike smoothing uses more which makes complete sense.  That's a different issue.

I also don't think it matters if we use all of the terms in a given order.  The polynomials are orthogonal.  There's really no reason to calculate terms if they will never be used.  I actually do use them all when looking at round robin results but the largest errors have always been in lower order terms so they really haven't changed the analysis.

Also DFTF might run noticeably faster if we cut back how many Zernike terms are included in the best fit calculations.

I don't particularly want to change how DFTF looks to the user.  I really just want to fix this one bug related to regions. :smile:

githubdoe:
Sorry I'm not questioning your understanding of the indexing.  I'm saying I don't trust that I always got that right.  Especially since you talked that I don't display all or calculate all.  The reason I use that many terms is that 36 is way to small to capture some of the obvious ringed zones that ATMs tend to make.  It showed a smoothing that was not realistic at all.   That was more important when creating surfaces from fringe tracing.  However now with DFT analysis those upper zernike terms don't matter much since we don't use them but just display them.

githubdoe:
Mike Peck explained why you must use all or none of a given order several years ago on the group.  I had thought the same as you that it did not matter.  But that is not the case when computing the terms.  It does not matter if you don't use all of the orders of course but you must use the complete set inside the orders you do include.

I think that is the reason you will still see a computation for 49 but that I don't use it.  It is part of the next order.  I deleted all the other partial terms but was afraid I had an off by one error so kept the 49th around in case I needed it.  But forgot about it in the long run.

gr5:
It's only m_norms that is calculated to one place too many which is a trivially fast calculation.

So I'm a bit confused in how many terms are in each order.  I thought order 12 is 49 terms (7^2) and order 10 is 36 terms (6^2) and so on in both directions.  That's what the zernike smoothing claims.  But maybe zernike smoothing should be using one less zernike term?  Or maybe it already is.  I really didn't look much at that code.

So I'm not sure if you should reduce m_norms by one or increase the zernike terms by 1.

githubdoe:
How many terms are in each order is computed by code provided by Mike Peck.   I use it in the smoothing to compute the numbers shown in the dialog.  So somewhere in that code you can find the formula.  It is not a simple square IIRC.

gr5:
You take the order, 12.  Divide by 2.  Add one.  Square it.  That's the formula you used.  so (12/2)+1 is 7 and you get 49.  So all should be 49.  That's what I located in your code and that's also what the zernike smoothing dialog displays for terms versus order.

So it sounds to me like I should try increasing Z_TERMS to 49.

githubdoe:
Yes it does does.  Good thing it does not really matter to the results.

atsju (commented):
I'm far from expert but from your discussions and my understanding, I also agree Z_TERMS should probably be 49.  
The correction of the loop is of course still valid change.  

I noticed in `wftStats::computeZernStats`, there is one loop that goes to `Z_TERMS-1` and the loop index is touched inside the loop like to ignore first term after spherical. Strange enough to have a check.

zernikeprocess.cpp
```diff
@@ -861,7 +858,7 @@ void zernikeProcess::fillVoid(wavefront &wf){
                         zpolar.init(rho,theta);
                         double v = 0.;
 
-                        for (int z = 0; z < m_norms.size(); ++z){
+                        for (int z = 0; z < wf.InputZerns.size(); ++z){
```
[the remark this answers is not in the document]
gr5: will do.

gr5:
Well, as I suspected, simply changing Z_TERMS to 49 adds another Zernike term to the display.

Dale, are you sure you want to change the behavior of DFTFringe for this one?  This started out as a simple off-by-one error where we were going beyond the end of an array because we were checking a different array for the size.  I fixed that but now you want me to mess with Z_TERMS as well?  I'm fine with it if that's what you want.

Let me know.  If just fixing the bug is enough then you can approve and merge this version.  If you want the change to Z_TERMS, I'll commit and push that change.

Also I did just now fix the warning message by using size_t.

Capture

zernikeprocess.cpp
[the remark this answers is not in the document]
gr5: fixed.  pushed.

gr5:
Also if Michael Peck is right about "solve" working better with a complete order of terms then we should change Z_TERMS to 49.  Let me know if that's what you want.

gr5:
Just to be clear: Z_TERMS doesn't just change the display.  It also changes the number of zterms going into the solve.  The two things match - all terms calculated are displayed.

githubdoe:
I think we do want that term.  But check the results of RMS to see if it makes any difference.  I expect it does not.  If not then lets add it.

githubdoe:
I found that all places that iterate using Z_TERMS use the < operator.  Which then means Z_TERMS is really +1 the number of terms desired.  However there is one place that compares the loop against Z_TERMS -1.  That could be a bug.  In wftstats.cpp

void wftStats::computeZernStats( int ndx){

    zNames.clear();
    zerns.clear();
    zernMinMax.clear();
    zernStd.clear();
    int col = 4;
    for (int zern = 4; zern < Z_TERMS-1; ++zern){

gr5:
I'll look at wftStats later. Here is what I found out with 49 versus 48 terms.

Using DFTF49 (49 terms) I processed an igram, then simulated an igram copying the zernike terms but the 49th term I set to 0.2 waves (yes the simulation also has a new 49th term called "5th Spherical").

I checked total RMS and Strehl and both had the same values.  I did not expect this.  I thought RMS was calculated based on all the error.  Not just the errors from the Zernike terms?  I thought the RMS error would be much worse with the extra 0.2 waves of error but RMS was the same at .125.

That was my first surprise.

I saved both wavefronts (one derived from an igram and one simulated) and also saved to zerns.csv.  The Zerns where very very close - the worst ones I think are .005 waves different.  I provide a link below to the spreadsheet.  I'm talking about the 3rd row of the spreadsheet which shows the differences.


Step 2, I ran DFTF48, loaded both wavefronts.  No changes to RMS error or Strehl for either waveform.

I then saved these 2 to zerns.csv.

The "real" wavefronts matched extremely well. (row 11 in spreadsheet).  Good news.  Basically having 48 versus 49 terms doesn't change things much unless...

Second surprise - on the simulated wave with the 0.2 wave error in the 49th term:
although most of the zernikes from the second set agreed with the first set, it had significant spherical errors in all the spherical terms.  Nothing too serious but the worst one was .034 waves off.  These I highlighted in **yellow** (row 12 in spreadsheet).

Conclusion: Mike Peck is right - it really does matter - at least if there are large errors in the 49th term (or probably any spherical term).  If not then it doesn't really matter.  

I suppose we should probably keep this as Z_TERMS set to 49.

**Please look at the small spreadsheet,** and let me know your thoughts.  I'll look at wftStats next.  It might not be today.  I don't know why but to get this next link to work I had to copy and paste it into my browsers URL bar.

http://gr5.org/zerns49.xls

githubdoe:
You should not have been surprised.  I have stated several times here and on the group that RMS is derived from the wave front itself and not the zernike values.  The zernike values are only used to remove the unwanted terms (tilt, defocus, SA  from testing at ROC)  from the raw wave front.
Of course Strehl is computed from RMS so it will not change either.

githubdoe:
I think one way to look at it is
Adding Zernike values lets it lower the RMS of each of the terms needed by added a little to the additional terms.  The overall wave front RMS will still be the same (I think)  just  with the extra terms those new terms can take up some of the burden.

githubdoe:
When it comes to math stuff Mike Peck is always correct.  It probably would be possible to find that conversation back in the archives.  Searching for something like "extra spherical terms."  There IIRC he explained why what I did to begin with should not be done.

gr5:
You misunderstand.  You just confirmed that RMS is calculated the way I thought HOWEVER

When I took the existing waveform AND ADDED MORE ERROR, the RMS didn't change.  I took the existing wavefront, and then used that to simulate a wavefront.  The simulated wavefront however had an extra 0.2 waves of error.

I guess what happened is that the simulated wavefront was basically very much smoothed so that would have lowered the RMS error and then I coincidentally added just the right amount of error to bring the RMS back up again.  I'll test that again to confirm.

gr5:
Yeah that's what it was.  Didn't think of that.  So just now I repeated teh experiment but just used the zernike terms exactly from the first igram.  And the simulated had lower RMS error (.110 versus .125).  When I simulated again but set the 49th term to 0.2 waves, the RMS error went back up to .125.  So I should have just changed one variable at a time.

So really the only surprise was how much the "solve()" function gets different results depending on how many zernike terms you include.  I guess that makes sense to me now.

So are you still going to be okay with adding one more term (49 terms instead of 48)?  People may complain that some of their zernike values (the spherical ones) come out a little different.  Probably not though.  Unless they have a *lot* of 5th spherical error.

githubdoe:
Ok,  I think I understand.  Yes I think we should include that term.  It will change any zernike table .csv that people write out however.  Perhaps we alert them in the release notes.    Perhaps also discuss it on the group to see if Mike and others think it is a good idea to update that.    If you want I will write up a summary and post it there for comment eventually.

gr5:
So far I noticed that it adds one more zernike to:
zerns.csv output
left side of gui
igram/wavefront simulation dialog

I feel like I've seen it at least one more place.  Basically anywhere there were 48 zernike's there are now 49.  Those are the 3 most likely places for people to notice it.

It shouldn't affect the first 48 zernike values at all for most mirrors but could affect spherical aberration values (including S.A.).

gr5:
Regarding the code in wftStats, somehow I never stumbled onto that feature before despite trying to go through every menu item in the past.  I guess I mostly concentrated on files and tools menu items.  Anyway I'm still testing it out.  Nice feature.  I don't think it's going to matter so I think I will remove the "- 1" for consistency but for the "box" graph (is that what they call it?) which shows a box representing the minimum, maximum, etc for each zernike value, on my screen it only goes up to 38 values.  I haven't figured out why but I think it is related to my screen resolution maybe?  That doesn't make sense either.  Anyway I'm going to put in a breakpoint.  I think it calculates the values for each BOX up to 47 terms (old code with Z_TERMS=48) but then only has room to display the first 38.  I don't think it will hurt to get rid of the "- 1" but still testing.

I hurt my back yesterday and sitting in front of the computer was too much last night and had to quit early.

gr5:
I left the "-1" in wftstats because there is this code where it references zern+1.  It doesn't matter much anyway because as far as I can tell it only goes out to 38 terms displayed.  At least on my monitor.  And in the pdf and png.

I could make the for loop more complicated where it uses a limit of zern < Z_TERMS for non-pairs of terms and uses zern < Z_TERMS-1  for paired terms.  But since it only displays up to zterm 37 I don't think it matters.

https://github.com/githubdoe/DFTFringe/blob/e541c41d4f00bd4a8ee88a775987c51a09669563/wftstats.cpp#L143-L144

gr5:
So I have now checked in the latest code with all discussion included.  After all that the changes are pretty simple to read through (files tab above).  I changed Z_TERMS to 49.  It seems to work great.  I've done quite a bit of testing.  

Dale, Please comment or otherwise please approve and merge, thanks.

zernikedlg.h
```diff
@@ -27,7 +27,7 @@
 #include <vector>
 #include <QVector>
 
-#define Z_TERMS 48
+#define Z_TERMS 49
```
atsju, line 30:
```suggestion
#define Z_ORDER 7
#define Z_TERMS (Z_ORDER*Z_ORDER) 
```
I suggest this for code readability and prevent futur errors. Feel free to adjust. 
Comment is also welcome. For example to remember that all terms of a given order should be used.

githubdoe (approved):

zernikedlg.h
gr5: Oops.  just merged.

It's more complicated.  Order is 12 (not 7).  It would be like this:

#define Z_TERMS (Z_ORDER/2+1)*(Z_ORDER/2+1)


Note that orders are always even numbers.  Z_TERMS are always perfect squares

Maybe just add a comment with the math right there at the Z_TERMS code?

atsju: There are several places where this math are done. Ideally there would be some helper function and necessary defines somewhere.  
Let's just add a comment here to start with

PR #93, opened 2023-09-17
[bug] fix windows not closing on application close

atsju:
I fixed the "remove test stand astig" windows not closing along with memory leaking.
Note I used https://doc.qt.io/qt-5/qpointer.html which is really great for this purpose. It is used nowhere else in DFTFringe so it's worse mentioning it exists.

mention #82

### DIFF
--- a/foucaultview.cpp
+++ b/foucaultview.cpp
@@ -38,8 +38,12 @@ foucaultView::foucaultView(QWidget *parent, SurfaceManager *sm) :
 
 
 foucaultView *foucaultView::get_Instance(SurfaceManager *sm){
-    static foucaultView m_instance{0, sm};
-    return &m_instance;
+    //static foucaultView m_instance{0, sm};
+    //return &m_instance;
+    // Take care. This is non standard init for when the singleton is supposed to be deleted by parent
+    // keeping original version will call class destructor and on_exit will try to clean up static variable m_instance. But the instance doesn't exist anymore.
+    static foucaultView *m_instance = new foucaultView(0, sm);
+    return m_instance;
 }
 
 foucaultView::~foucaultView()
--- a/foucaultview.cpp
+++ b/foucaultview.cpp
@@ -7,6 +7,7 @@
 #include <QVector>
 #include <QMenu>
 #include "zernikeprocess.h"
+#include "spdlog/spdlog.h"
 
 extern double outputLambda;
 
@@ -49,7 +50,7 @@ foucaultView *foucaultView::get_Instance(SurfaceManager *sm){
 foucaultView::~foucaultView()
 {
     delete ui;
-    qDebug() << "foucaultView::~foucaultView";
+    spdlog::get("logger")->trace("foucaultView::~foucaultView");
 }
 QString getSaveFileName(QString type){
     QSettings settings;
--- a/foucaultview.cpp
+++ b/foucaultview.cpp
@@ -9,7 +9,6 @@
 #include "zernikeprocess.h"
 
 extern double outputLambda;
-foucaultView *foucaultView::m_instance = 0;
 
 foucaultView::foucaultView(QWidget *parent, SurfaceManager *sm) :
     QWidget(parent),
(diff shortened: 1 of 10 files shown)


atsju:
note to myslef: probably makes the qwizard modal. To be tested and must be reworked.

atsju:
@gr5  ready for review (no pressure for timeline).
I will do them one by one because doing it the right way and ensuring that everything is cleaned correctly with Valgrind is not so easy.

foucaultview.cpp
```diff
@@ -39,15 +39,18 @@ foucaultView::foucaultView(QWidget *parent, SurfaceManager *sm) :
...
 
 foucaultView *foucaultView::get_Instance(SurfaceManager *sm){
-    if (m_instance == 0){
-        m_instance = new foucaultView(0,sm);
-    }
+    //static foucaultView m_instance{0, sm};
+    //return &m_instance;
+    // Take care. This is non standard init for when the singleton is supposed to be deleted by parent
+    // keeping original version will call class destructor and on_exit will try to clean up static variable m_instance. But the instance doesn't exist anymore.
+    static foucaultView *m_instance = new foucaultView(0, sm);
```
gr5, line 46: This looks quite wrong to me.  Every time get_Instance is called a new class is created.  Whereas in the old implementation it only created a new foucaultView class if it hadn't been created yet.  Maybe the above line of code should be called outside of any function?  Something like Dale did on line 12 in the old code?

atsju: It's a static variable inside of a function. Here static has meaning of one assignation over whole program. First call will assign. Subsequent call will reuse same variable without assignement. And it will be clearly deleted on exit (well here with a pointer it will not but it's done on purpose as described in comment) . 

I will triple check tomorrow but here is doc
https://en.cppreference.com/w/cpp/keyword/static
Static has many meanings in CPP. Here we are in second case https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables

> Variables declared at block scope with the specifier static or thread_local (since C++11) have static or thread (since C++11) storage duration but are initialized the first time control passes through their declaration (unless their initialization is [zero-](https://en.cppreference.com/w/cpp/language/zero_initialization) or [constant-initialization](https://en.cppreference.com/w/cpp/language/constant_initialization), which can be performed before the block is first entered). On all further calls, the declaration is skipped.

atsju: This is actually a very clean way to implement a Singleton. Everything gets cleaned up when it needs to be. It doesn't rely on OS to clean up missed things during exit as some of Dales Singletons did. I mentionned my Singleton design source in a previous PR. 

Some (probablt most, maybe all) Singletons should not even exist in DFTF architecture. It doesn't fit the architecture and especially doesn't mix well with QT parenting. It's just a "convenient" way to find an object without passing it everywhere or having it global. As it works I won't remove them. However I change their implementation because it makes valgrind analysis clean and thus helps to find the "real" leaks

gr5: Okay.  It's fine.  Nevermind.  I thought it was a member variable.  Dale had it as a static member variable which is similar but different.  I think variables have to be initialized before runtime.  No, instead this is a variable within a function.  A static variable.  It's local to this function but because it is static, it doesn't get deleted when you exit the function.  So how does it get deleted?  That's what I will look at next.

gr5: And valgrind is okay with this?  This doesn't get deleted in any destructor.  I mean I would have fixed Dale's code by having some destructor (maybe mainWindow?) call the destructor directly.  As long as valgrind is happy, I'm happy.

atsju: If it was not a pointer it would get destructed at exit. Static variables are destructed correctly. With pointer of course it does not get deleted by the Singleton class. I made it a pointer on purpose because this is a Qobject that has a parent. It is deleted/destructed by its parent when application closes. Double destruction was not ok. You cannot mix qt object ownership with own code ownership.

gr5: Okay.  I understand now.

gr5: I plan to look over the code again today.

surfacemanager.cpp
```diff
@@ -2583,8 +2590,17 @@ void SurfaceManager::computeStandAstig(define_input *wizPage, QList<rotationDef
...
-    standAstigWizard *wiz = new standAstigWizard(this);
-    wiz->show();
+    if(m_standAstigWizard == nullptr){
+        spdlog::get("logger")->trace("new standAstigWizard");
+        m_standAstigWizard = new standAstigWizard(this);
+        m_standAstigWizard->setAttribute( Qt::WA_DeleteOnClose, true );
+        m_standAstigWizard->show();
+    }
+    else{
+        // bring to front the already oppened window
```
gr5, line 2600: Excellent addition.

gr5 (approved):

PR #95, opened 2023-09-30
close camera calibration at end of application

atsju:
mention #82

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -281,7 +281,7 @@ void MainWindow::openWaveFrontonInit(QStringList args){
         if (pd.wasCanceled())
             break;
 
-        if (arg.toUpper().endsWith(".WFT")){
+        if (arg.endsWith(".wft", Qt::CaseInsensitive)){
             pd.setLabelText(arg);
             try {
             m_surfaceManager->loadWavefront(arg);
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -59,7 +59,7 @@ MainWindow *MainWindow::me = 0;
 MainWindow::MainWindow(QWidget *parent) :
     QMainWindow(parent),
     ui(new Ui::MainWindow),m_showChannels(false), m_showIntensity(false),m_inBatch(false),m_OutlineDoneInBatch(false),
-    m_batchMakeSurfaceReady(false), m_astigStatsDlg(0)
+    m_batchMakeSurfaceReady(false), m_astigStatsDlg(0), m_cameraCalibWizard(nullptr)
 {
     ui->setupUi(this);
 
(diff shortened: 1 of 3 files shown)


githubdoe (approved):
Just for information.  Camera Calibration was never used by anyone as far as I can tell.  It also was never verified that it worked correctly.  I got confused by different nomenclatures and was not able to compare with an example I found on the net that I had planned to use for verification.  It probably could be removed.  I hated to remove it because it was a lot of work and research that might be useful.  However it probably requires a large learning curve to actual be used by anyone.  I have abandoned all thought of ever using it.  Most camera and lens combinations we ATM's used in the past did not create distorted views enough to matter.

gr5:
I tried to use it but it wouldn't recognize the pattern.  Let me clarify:
Step 1: create pattern.  Works great.
Step 2: photograph pattern with camera and lens.  Works great.
Step 3: import photograph and calculate distortion of photographed pattern.  Could not get it to work.

As part of step 2, supposedly you should also use a specific "stop" device but I never got that far.

githubdoe:
I remember that could happen sometimes.  But no longer remember the details.  There were/are HTML links in the code comments or help file I think pointing to where I got the algorithms and web sites.  IIRC.

githubdoe:
I also remember that I was planning on validating it against another app where one could add distortion to a photo.  I would add the distortion to the grid pattern.  However the units and names it called the distortion did not match with any of the names or units of the calibration constants my app produced.  So I could not compare them directly and did not know how to reconcile that issue.  Thus there was no validation of my algorithm.

mainwindow.cpp
```diff
@@ -282,7 +281,7 @@ void MainWindow::openWaveFrontonInit(QStringList args){
         if (pd.wasCanceled())
             break;
 
-        if (arg.toUpper().endsWith(".WFT")){
+        if (arg.endsWith(".wft", Qt::CaseInsensitive)){
```
gr5, line 284: I'd really rather you don't change things like this.  The code is about equally easy to read either way.  The code does "the same thing" either way.  There is a tiny (very tiny) chance you will add a bug anytime you change *anything*.  And we don't need to worry about efficiencies.  If this code looped 10,000 times, maybe?  Certainly if it looped a billion times.  Some people might find the old code easier to read.

Having said all that I'd prefer you to leave it this new way as I plan to build and test this in the next few minutes and merge it.

gr5: Or was there a clangd/warning about the old line of code?

githubdoe:
That is a change that I like and would have done myself at the time had I knew about that option.

githubdoe:
However it does not make a difference for .wft files since this is the program that creates the extension.

PR #96, opened 2023-10-01
mirror dlg closing at end of application

atsju:
Also improved some code from #88

mention #82

### DIFF
--- a/mirrordlg.cpp
+++ b/mirrordlg.cpp
@@ -17,19 +17,19 @@
 ****************************************************************************/
 #include "mirrordlg.h"
 #include "ui_mirrordlg.h"
+#include "spdlog/spdlog.h"
 #include <QtGui>
 #include <QFileDialog>
 #include <iostream>
 #include <fstream>
 #include <QMessageBox>
 #include <QDebug>
+
 QString mirrorDlg::m_projectPath = "";
-mirrorDlg *mirrorDlg::m_Instance = 0;
+
 mirrorDlg *mirrorDlg::get_Instance(){
-    if (m_Instance == 0){
-        m_Instance = new mirrorDlg();
-    }
-    return m_Instance;
+    static mirrorDlg m_Instance{};
+    return &m_Instance;
 }
 
 mirrorDlg::mirrorDlg(QWidget *parent) :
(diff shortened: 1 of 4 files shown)


githubdoe (approved):

mirrordlg.cpp
```diff
@@ -17,19 +17,19 @@
...
+
 QString mirrorDlg::m_projectPath = "";
-mirrorDlg *mirrorDlg::m_Instance = 0;
+
 mirrorDlg *mirrorDlg::get_Instance(){
-    if (m_Instance == 0){
-        m_Instance = new mirrorDlg();
-    }
-    return m_Instance;
+    static mirrorDlg m_Instance{};
```
gr5, line 31:
```suggestion
    static mirrorDlg m_Instance;
```
Why the {} initializing?  I believe we're just using the default constructor here.  Nothing to pass in.

atsju: To be honest I copied syntax from https://leimao.github.io/blog/CPP-Singleton/ see "Scott Meyers Implementation" part. All other singleton I migrated have this syntax and this is probably my main argument (+ matter of taste).

I also checked on isocpp website (Have a look but it's very very long. They try to have recommendation to write best possible C++ and the recommendation are evolving over time)
It seems they recommend the syntax with {} https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#es23-prefer-the--initializer-syntax but I have also seen some of their examples that use your syntax for simple objects...

So feel free to commit your suggestion or not before merging. Both are OK.

gr5: I'm on the fence so since Dale didn't chime in I'll leave it as it is.  It's a new syntax that came in with C++11 so I rarely see it and Dale might not be familiar with it.  Newer recommendations aren't always better.  Sometimes it's better to have a more consistent style.

I guess I'll leave it with your new style for now.  Especially since it's "new" code.  Let's not change every instance where we create a class variable and instantiate it as that would be dangerous but I guess it's fine for new code.

githubdoe:
Yes I'm not familiar with that curly brace syntax either.  But OK.

PR #102, opened 2023-10-15
fix preference dialog closing

atsju:
mention #82

### DIFF
--- a/lensetablemodel.cpp
+++ b/lensetablemodel.cpp
@@ -6,7 +6,6 @@ LenseTableModel::LenseTableModel(QObject *parent)
 {
 }
 
-
 QVariant LenseTableModel::headerData(int section, Qt::Orientation orientation, int role) const
 {
     if (role == Qt::DisplayRole)
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -319,6 +319,7 @@ MainWindow::~MainWindow()
     delete m_ogl;
     userMapDlg->close();
     mirrorDlg::get_Instance()->close();
+    settingsDlg->close();
     delete ui;
 }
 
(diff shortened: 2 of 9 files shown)


gr5:
Getting "segmentation fault" when I try to close the application with the preferences window open.  If I open the application and immediately close it, no seg fault.  If I open the app, open "config" "preferences", then close app, I get seg fault.  settings2 destructor calls settingsIgram destructor which gets a seg fault.

I'm thinking this is an old bug that is only now showing up?  Not sure.

Screenshot 2023-10-15 193025

atsju:
This is pretty much all linked to the Singleton thing not managing destruction yes. We should really get rid of them but that's a big work with associated risks... 

Sorry I missed the fault, I didn't retry on windows. Or maybe I just missed it.

atsju:
I finally had a look.
This problem is already present in origin/master.
It will happen is preferenceDlg has been opened at least once (it can be already closed or closed after main windows) and create a segfault on windows (but not on Linux I think).
It might have to do with singleton, but I don't see where right now. For once it's doesn't seem to be linked to a double destruction (once by singleton and once by QT parent/child relations) like the other cases I had.

I will continue investigation but this might be merged as is if you want. Indeed, the user experience is better and even if the code execution is bad at close, it's not worse than what it already is.
Of course I prefer to fix it, it's just I don't know when I will be able to do it.

atsju:
@gr5  looks good now. 
Tested on windows and linux

gr5:
Thanks, hopefully I'll look at it today.  I have 5 minutes now but that's not enough time.

gr5 (approved):
Works great!  Thanks for fixing the seg fault.

Issue #104, opened 2023-10-16
Remove Interpolation drop down from rotate wavefront dialog


gr5:
I actually already wrote the code and am creating a request to match with the pull request.

There were several options (linear, nearest neighbor, cubic, etc) that were ignored by the rotation code so I wanted to just get rid of this bit of GUI.  No need to waste people's time trying to choose which interpolation method when it is then ignored (Dale does bilinear interpolation with some custom features related to when a source pixel is masked).

gr5:
fixed in pr #105

PR #105, opened 2023-10-16
Implements issue #104 - removes "interpolation" drop down

gr5:
Because the results are ignored by the code anyway and always does a bilinear interpolation.


githubdoe (approved):

atsju (approved):

atsju:
#104

Issue #112, opened 2023-11-07
7.1.0 crashes when doing wavefront smoothing


gr5:
**Describe the bug**
Menu item "tools" "smooth current wave front" - if you increase size of zernike terms or just click "create wave front" button, it crashes

correction!  It only crashes if you have an inner outline:
After creating outer outline, but before processing the igram, select the "central hole" radio button near the top, then use the mouse to draw a circle inside your igram.  I did about 10% of the total diameter.  I've tried many different sized inner diameters including some only a few pixels across.

**version**
tag 7.1.0 (not released) - the version built by github.  
7.0.0 also crashes

When I build it myself in QT in debug or release mode it does not crash.  I have a different version of the armadillo .h files.

**stacktrace**
```
[2023-11-07 10:58:34.004] [critical] Unexpected issue. Stacktrace:

 2# 0x000000006FC56FA8 in C:\Users\gr770\DFTF710RC\libstdc++-6.dll
 3# 0x000000006FCFD093 in C:\Users\gr770\DFTF710RC\libstdc++-6.dll
 4# 0x000000006FD05E12 in C:\Users\gr770\DFTF710RC\libstdc++-6.dll
 5# 0x00000000005B9835 in C:\Users\gr770\DFTF710RC\DFTFringe.exe ...ingw32/8.1.0/include/c++/bits/basic_string.h:211
 6# 0x0000000000504FF4 in C:\Users\gr770\DFTF710RC\DFTFringe.exe include/armadillo_bits/Mat_meat.hpp:424
 7# 0x0000000000507AAE in C:\Users\gr770\DFTF710RC\DFTFringe.exe zernikeprocess.cpp:1422
 8# 0x00000000005085A0 in C:\Users\gr770\DFTF710RC\DFTFringe.exe
 9# 0x000000000050C4C2 in C:\Users\gr770\DFTF710RC\DFTFringe.exe
```

gr5:
I learned more. I didn't realize it but I had a small "inside outline" in the OLN file.  It was so small I didn't notice it.  This could be a very old bug.

atsju:
relevent version : cd52669dd953b2f5b01b3e1cb6947dddbfd0dff1
I just tried with this binary and did not reproduce : https://github.com/githubdoe/DFTFringe/actions/runs/6776965033

Please share the igram and oln you used. It hink you can zip it and put here

gr5:
You have to create an inner outline:
[...]

PR #114, opened 2023-11-08
Fixes issue #112 (smoothing wavefront crashes if obstruction)

gr5:
Disabled zapmC() which crashes.  It doesn't seem to crash when I build it in qt creator but the version built by github definitely consistently crashes.  Don't know why but probably some according to Dale, the code might have known bugs.


githubdoe:
The known bugs are not execution bugs but instead compute wrong values for the zernike terms.  IIRC

githubdoe (approved):

Issue #115, opened 2023-11-20
Edge Mask shouldn't be saved in OLN files


gr5:
**Is your feature request related to a problem? Please describe.**
When edge mask is non-zero and also unchecked, sometimes it seems to act as though it is checked.  The checkbox and value are found in "configuration" "mirror"

**Describe the solution you'd like**
Just set the value to zero when unchecked.

**Describe alternatives you've considered**
Instead, all code that uses edge mask could only use the value if checkbox is checked.

**Additional context**
https://groups.io/g/Interferometry/topic/102697035

githubdoe:
The edge mask is also saved in the .oln file.  I don't think that it should be stored there.  It is a value of the mirror configuration.  Did I ever think it should go into the .oln file?  I don't remember thinking that.

Also I prefer it to never set the value to zero if you uncheck it.  I like to be able to uncheck and the recheck it to see how the analysis is changed but don't want to have to reset the value.

githubdoe:
I reviewed my old code and I never saved the edge mask in the .oln file.  The more I think about it the more I'm against doing so.
It is a function of the mirror and not of the interferogam.

gr5:
I never purposely had it save to the oln file.  I look at the code and see when that change happened.

gr5:
Looks like you did it back in July of this year.

![Capture](https://github.com/githubdoe/DFTFringe/assets/5414808/df2e90fb-8e1c-4860-93fa-160d290fa9ef)

gr5:
See your commit comment highlighted in the upper part showing all the commits in blue.  Filename is on the right highlighted in blue.  Code that saves is circled in red.  There is more code to read it back in.

I like that it is in the OLN file because when you come back to this mirror to check the OLN file it seems it should load the config from the OLN file maybe?  I'm not 100% sure what the best process is for re-analyzing old igrams.  But it sure is convenient to have it in the OLN.

Ideally the OLN would point to an old config "ini" file and strongly suggest the user loads that into the mirror config as well.

I agree that it's basically associated with the config file because if you do it for one igram on a mirror you probably want to do it for all of them.

I suggest for now I simply fix it such that if the checkbox is unchecked it does not save this in the OLN file (just like if it's set to zero it does not save it).

gr5:
The registry currently stores *both* (and separately) the aperatureReduction[sic] value and if it is checked.

[...]

PR #116, opened 2023-11-22
Fixes issue #115 - issues with "edge mask" aka aperture reduction

gr5:
When reading old OLN files, ignore edge mask
When writing OLN files don't save edge mask
Fixed 2 locations where edge mask needs to be set to zero when checkbox is not checked in mirror dialog Fixed bug where edge mask is checked if not found in OLN file!

### DIFF
--- a/igramarea.cpp
+++ b/igramarea.cpp
@@ -152,6 +152,8 @@ void IgramArea::computeEdgeRadius(){
     mirrorDlg &md = *mirrorDlg::get_Instance();
     double pixelsPermm =(m_outside.m_radius/(md.diameter/2.));
     m_edgeMaskWidth = md.aperatureReduction * pixelsPermm;
+    if (md.m_aperatureReductionEnabled == false)
+        m_edgeMaskWidth = 0;
 
 }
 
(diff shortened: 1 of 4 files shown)


igramarea.cpp
```diff
@@ -2140,45 +2142,6 @@ void IgramArea::loadOutlineFile(QString fileName){
...
-                break;
-            case QMessageBox::No:
-            default: // if they click red X do same thing as "no" button
-                md.changeEdgeMaskvalues(md.aperatureReduction);
-                break;
-            }
-        }
-    }
-    else{ // just enable edge mask check box to use the current value.
-        md.changeEdgeMaskvalues(md.aperatureReduction);
```
gr5, line 2180: So this line of code is particularly nasty and found when reading OLN file: both the older V6 style OLN file and also when reading V7.  It checks the "edge mask" checkbox even if nothing is found regarding edge mask in the OLN file!

gr5:
I spent a lot of time testing all the changes.  It seems to work fine now.

githubdoe (approved):
Thanks.  Looks like it was pretty messed up with a lot of omissions.  Makes me wonder what I was thinking.  Guess I was distracted.  Did not realize all the implications.  Still puzzled why I did so much work in the reading and writing of it in the .oln file.  That worries me that I thought it necessary.  Now of course I think it was not.

githubdoe (approved):

gr5:
Dale, Do you want me to release 7.2 now just for this change?  It's not too much trouble.  I need to test the final release if I do a release it but I'll probably just do a few quick tests (load igram, process) as not much changed.  And I already tested the changes.  Although I will need to update the revision history and you'll need to approve that.

githubdoe:
I think it would be nice but I will leave it up to you to decide based on your schedule and such.

igramarea.cpp
```diff
@@ -2257,40 +2220,7 @@ void IgramArea::loadOutlineFileOldV6(QString fileName){
         mirrorDlg &md = *mirrorDlg::get_Instance();
```
atsju, line 2220: This line can be deleted as `md` is not used anymore. This will fix the `[-Wunused-variable]`

atsju:
> Dale, Do you want me to release 7.2 now just for this change? It's not too much trouble. I need to test the final release if I do a release it but I'll probably just do a few quick tests (load igram, process) as not much changed. And I already tested the changes. Although I will need to update the revision history and you'll need to approve that.

I wonder if this should be v7.1.2. This would be a pure bugfix release. 
Sorry if this seems futile and great work to find and fix this. Thank you

igramarea.cpp
gr5: oops.  Already merged.  Didn't see your comment until now.

atsju: you can fix in the other PR. It's really minor. 
This is how dead code comes to live :ghost:

PR #128, opened 2024-02-27
fix GH action doing the releases

atsju:


atsju:
@gr5 merge this and redo the tag should work.
As the release has not been done and no code modification I think it's safe to delete the v7.2.0 tag to move it 1 commit further.

gr5 (approved):

PR #130, opened 2024-02-27
update armadillo to v12.6.7

### LINKED ISSUE #129: github build DFTFringe crashes when calling zapm
current version of DFTFringe in master (tagged V7.2.0 at the moment - the commit merged in today 2/27/2024) is crashing when built with github.  But works fine when I build it on my computer with QT (both release and debug).  The good news is that it is consistent and happens with only seconds of usage.  I either simulate an annular igram or analyze an annular igram.  simulate is faster to get to the crash for me.

It's crashing on this line of code in initgrid()
https://github.com/githubdoe/DFTFringe/blob/094eca98b0dae466022827cd26d6a24adae74797/zernikeprocess.cpp#L1513

It never gets to zapm() but crashes when creating the parameters to zapm().  Within armadillo I think.

I need to do other things this afternoon but I'm going to try using armadillo 12.6.1 (Dale and I use I think 12.6.7) and see if it crashes.  If it does, then that's good (I guess?) and we can change the build process.

atsju:
- Fixed Lapack build to generate Lapack and Blas DLLs instead static libs. See https://github.com/Reference-LAPACK/lapack/issues/305 where I added a comment for clean fix.
- Groupped Lapack and Armadillo build so that Armadillo is configured correctly and makes use of Lapack+Blas
- Add define of ARMA_DONT_USE_WRAPPER to fix linking process when using Armadillo
- Update Armadillo version to 12.6.7

Closes #129

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -8,7 +8,7 @@ on:
   workflow_call:
 
 env:
-  armadillo_version: 12.6.1
+  armadillo_version: 12.6.7
   QWT_version: 6.1.6
   openCV_version: 4.6.0
   QT_version: 5.15.2
@@ -160,7 +160,6 @@ jobs:
         uses: actions/cache@v4
         with:
           path: |
-            build_armadillo\libarmadillo.dll
             build_armadillo\tmp\include
           key: ${{ runner.os }}-armadillo-${{env.armadillo_version}}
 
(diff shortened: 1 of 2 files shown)


atsju:
@gr5 you can test if [this](https://github.com/githubdoe/DFTFringe/actions/runs/8069028026/artifacts/1279692747) crashes

gr5:
That also crashes.  :frowning_face: 

I guess I'll pull this version anyway.

atsju:
@gr5 This is ready for review. Please check the build and merge.

I recommend you update release note and tag 7.2.1 (let the tag 7.2.0 alone. There is no associated release and executable so it's OK)

gr5:
Thank you so so much!  I will review, merge, test late tonight and hopefully release tonight as well.

gr5 (approved):
changes look good.  Used installer and dftf seems to be working!  Will test final release version more thoroughly.

PR #145, opened 2024-03-08
Fixes a crash for when you switch from annulus to regular and back again

### LINKED ISSUE #144: another out-of-matrix crash
**To Reproduce**

This is extremely repeatable.

open DFTF, in mirror config click annular box and set to 50%.  Nothing else matters in mirror config.
simulate igram, settings don't matter but I did 600 pixels, click ok
in igram window click "done" (no need to set inner outline), click "compute surface"

now go back to "mirror config".  Uncheck annulus, click ok.  Wait a few seconds for it to process.  now go to mirror dialog, check annulus and click ok. 

You can see the progress bar slide across and it crashes part way through - here's the line of code where it crashes in ZernFitWavefront()

For my 600X600 igram and DFT, it crashes when i reaches 211288.  m_rhoTheta.n_cols is unfortunately larger at 277976.
Capture

gr5:
Fixes #144 

I almost made it set m_lastusedAnnulus to false but that would also be bed as initgrid still needs to recalculate zerns if called while mirror is non-annulus.

This should fix the problem.  I mean it passes my test.  If approved I will create a release and do more testing of annuluar and normal igrams.

### DIFF
--- a/zernikeprocess.cpp
+++ b/zernikeprocess.cpp
@@ -766,6 +766,8 @@ cv::Mat zernikeProcess::null_unwrapped(wavefront&wf, std::vector<double> zerns, 
         // annular wave fronts already have this made elsewhere.
         if (!md->m_useAnnular){
             m_rhoTheta = rhotheta(nx ,wf.m_outside.m_radius, midx,midy, &wf);
+            if (m_lastusedAnnulus)
+                m_needsInit=true;
         }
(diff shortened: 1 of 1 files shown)


githubdoe (approved):
I'm going to trust your fix without thinking about it much.  I don't have time to consider it at the moment.   If it works going from and to an annulus then good.

PR #148, opened 2024-03-12
No need to copy any DLLs manually for debug anymore

### LINKED ISSUE #146: Check how DLLs are added in .pro
Sounds strange to me that the DLLs needed to run DFTFringe when it has just been build in Qt Creator need to be manually copied. Something must be off in the process.

I had a look at [Qt documentation](https://doc.qt.io/qtcreator/creator-project-qmake-libraries.html) and here is what I get when following the recommended process.

```
win32:CONFIG(release, debug|release): LIBS += -L$$PWD/../build_openCV/install/x64/mingw/lib/ -llibopencv_calib3d460.dll
else:win32:CONFIG(debug, debug|release): LIBS += -L$$PWD/../build_openCV/install/x64/mingw/lib/ -llibopencv_calib3d460.dlld
else:unix: LIBS += -L$$PWD/../build_openCV/install/x64/mingw/lib/ -llibopencv_calib3d460.dll

INCLUDEPATH += $$PWD/../build_openCV/install/include
DEPENDPATH += $$PWD/../build_openCV/install/include
```

I think we must update the .pro to take DLLs cleanly. It might even help for creating the installer.
I let this here for tracking and will confirm+solve later.

atsju:
fix #146

### DIFF
--- a/DFTFringe.pro
+++ b/DFTFringe.pro
@@ -33,9 +33,9 @@ win32 {
     message("Using WINDOWS project configuration.")
 
     CONFIG( debug, debug|release ) {
-        LIBS += ..\qwt-6.1.6\lib\qwtd.dll # debug
+        LIBS += -L..\qwt-6.1.6\lib -lqwtd # debug
     } else {
-        LIBS += ..\qwt-6.1.6\lib\qwt.dll # release
+        LIBS += -L..\qwt-6.1.6\lib -lqwt # release
         CONFIG+=force_debug_info # keep debug infos (even in release build) to be able to link stacktrace address to actual function
         CONFIG+=separate_debug_info # separate debug infos into a .exe.debug to not grow the .exe
     }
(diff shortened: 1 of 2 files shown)


atsju:
Please give the new DFTFringe.pro a try. With these changes, it's not required to copy manually any DLL into build folder anymore as they are already found inside their folder.

atsju:
> I don't remember having to do that (copy dlls to build folder) but you could be right. I'll investigate.
> 
> I thought the dlls are already pointed to in the project file and I thought maybe the IDE takes care of copying them? I really don't remember so I'll look into this.

What happens if you delete you build folder and do a clean build ? Personally, I required some DLLs to be in the build folder.  
Maybe you added te necessary folders to your windows path to avoid this ? Works also but pollutes the path with DLLs. Not so clean if you build different tools with different dependencies on same computer.

githubdoe:
For george and me who use Qt creator builds.    We have path variables to the non-Qt dlls.  Well at least I do.  Then I think we have path variables to Qt libs that get us all those one users of the software need.  So our build directories don't have any .dlls in them at all I think.

atsju:
> For george and me who use Qt creator builds. We have path variables to the non-Qt dlls. Well at least I do. Then I think we have path variables to Qt libs that get us all those one users of the software need. So our build directories don't have any .dlls in them at all I think.

I also use Qt Creator but didn't add the DLLs to Windows path. 
Well this modification has no impact for you then. But it eases the process for clean machines.

gr5 (approved):

gr5:
It turns out I *did* copy the dlls to the debug and release folders but I did it so long ago that I had forgotten.

gr5:
Julien - go ahead and merge this.  I approved it, I thought I'd let you do the honors with your new rights. :smile:

Issue #147, opened 2024-03-12
mirror config bugs


gr5:
**To Reproduce**
In mirror config make sure "units in mm" is checked, then load a json file version of an existing mirror config preferably with annulus enabled in that file.

  note that after loading, values are in mm but the "units in mm" box got unchecked.   It shows mm units but checkbox is cleared.

Now if you check the box things go weird and in particular the annulus diameter gets 7X too large.

**Expected behavior**
units shouldn't change.  Values loaded should match state of "units in mm" box.

**Desktop (please complete the following information):**
dftf 7.2.1
windows

githubdoe:
The bug is that .json code is reading the units value from the .json file.  However old config never read units config from the file but instead always expects units to be stored in mm and never in inches.  The .json read code should be altered to just set the units variable and check box to true.  Making sure first that writing .json file also stores  units in mm.

githubdoe:
I think I fixed it and did a push.

gr5:
1) you didn't have to fix it.  I could have done it.  Just letting you know.
2) I looked at your code.  Looks good.  I'll test it.
3) next time please push your changes in a differently named branch so we can do the whole pull request process.

atsju:
Let me know if you want me to cherry pick or rebase Dale commit in a fresh branch.

gr5:
Your fix works fine however...

I should have listed the other bugs but I was going to investigate them and fix them at the same time.  I said "bugs" for a reason.  but I should have explained that.

Anyway there are 2 more bugs.  

One is that annulus diameter caps out at 1 meter no mater what.  Someone might want to use this software on a 2 meter f/5 mirror with 51% annulus but dftfringe refuses.

Another is that if you are in inches mode when you read in the file, the annulus diameter appears to be sometimes the inches value and other times 4X too large.  Now I can't repeat quite exactly.  When you switch to inches mode, it actually switches the diameter to mm mode (if it had been inches but not if it had been mm).  Then when you switch to mm again it multiplies by *another* 25.4 and clips at 1000mm.  

I think I got the description close enough to repeat the bug - basically load file while in inches mode and then toggle mm/inches a few times.  This whole thing is weird because the annulus diameter isn't stored in the json file but seems to be calculated on load.  Which is a good thing but should prevent bugs like this.

[...]

PR #154, opened 2024-07-07
Fix Issue147 - problems with save/load mirror configuration json file

gr5:
One of the bugs had to do with a maximum value of 1000 for the annulus diameter.  I raised that to 10000 (10 meter mirrors max for hole in mirror) Another was that as you switched between inches and mm the annulus diameter value kept growing.

### DIFF
--- a/mirrordlg.cpp
+++ b/mirrordlg.cpp
@@ -277,8 +277,7 @@ void mirrorDlg::loadFile(QString & fileName){
 
         ui->name->setText(QJsonValue(loadDoc["name"]).toString());
         ui->unitsCB->setChecked(true);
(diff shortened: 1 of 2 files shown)


gr5:
Fixes issue #147

atsju:
> Fixes issue #147

Only for info : This won't auto close.
You need either "fix #147" or "fixes #147" and it must be written in a commit message or in the top comment/description of the PR. I edited it for you

atsju (approved):
Please merge this one first as the other contains changes from here also.

gr5:
Oops.  I forgot to checkout master before I did the "git checkout -b issue153".  Thanks for the warning.  Thanks for the help, both of you.

Issue #153, opened 2024-05-20
crash with defocus adjustement


atsju:
1. Create wavefront
2. use defocus adjustment (I use the box at the top with corresponding arrows) to add or remove some defocus
3. DFTFringe crashes

Maybe I did something I'm not supposed to do. I do not know this feature, juste wanted to play with it a bit

**Desktop (please complete the following information):**
 - OS: Win10
 - DFTFringe version: 7.2.1 official release

Here is the igram I use with corresponding outline. I reproduced with other ones too
[90deg.zip](https://github.com/githubdoe/DFTFringe/files/15375208/90deg.zip)

gr5:
This is probably the same crash in #151.  Almost certainly a duplicate of #151.

I've experienced this also.  If I don't use arrows and type in the values to reduce computation, I can try maybe 5 or 10 values before a crash.  If you just mess around it's pretty easy for me to make it crash within 10 seconds.

At the same time, it's pretty easy to get it to _not_ crash.  Be mindful of the values you enter.  Wait for the waveform to change before entering the next value.  Highlight only the digit you want to change and type the new value rather than deleting and then entering (which requires 2 recalculations).

If there was a recalculate button this could probably avoid the crash.

atsju:
Correct. Probably a duplicate. However only little defocus is needed to make it crash

gr5:
I'm going to look at this next.

gr5:
I found the bug.  Wasn't too hard although it took an hour or two.

The problem is in line 1279 below (circled in red).  When you move the defocus slider it instantly emits at least 2 events typically.  The m_row vector is cleared just above at 1273 and then when it processes events it calls this function again, adds in a lot of values and then returns and we add more values to m_row such that later m_row (and m_col) have double the size as desired.

Then it crashes in zernikeProcess.cpp line 779 because it uses m_row in the for(i...) loop and goes well past the end of m_rowTheta.

I'm not sure what the fix is.  Maybe disable this line of code if we are doing defocus.  I'm actually thinking this can cause other bugs as well - like if you click on "recompute" twice in a row quickly?  I tried that but it didn't crash.  Maybe it knows it doesn't need to call rhotheta().  But it seems like there could be other ways this could cause a crash.

![Capture](https://github.com/githubdoe/DFTFringe/assets/5414808/2e8fae89-64f4-470a-af0c-9a4cf7d97b39)

gr5:
[...]

PR #156, opened 2024-07-08
Fixes Issue153 - defocus adjustment crashes

gr5:
fixes issue #153 

This fix works great.  I had to disable a line of code and originally I was going to comment it out but decided to only disable the line for defocus (for now).

I think the line of code was meant to allow progress bars (a guess) but there are no progress bars visible at the time of this code running so I don't think it hurts anything.

I tested the hell out of this.  At first I got crashes after the 2 second "background" timer so I added the flag there as well.

@atsju or @githubdoe please review and approve


githubdoe (approved):

atsju:
Looks like you used same branch for both fix. So this branch contains #154 . 
I will have a look anyway

Issue #132, opened 2024-02-29
Saving on network drive is incredibly slow


atsju:
**Describe the bug**
As mentioned [here](https://github.com/githubdoe/DFTFringe/issues/98#issuecomment-1762958562) and in #131 DFTFringe is incredibly slow when saving on network drive

githubdoe:
I find many posts about that problem which is more than likely a Qt problem.  I know there are other apps based on Qt that I can't imagine have that problem.  So there must be away around it.  They say it is caused because Qt breaks the write up into very tiny byte steams.  That all I know at this point.  I realized I do have a way to test it here with a networked drive and it is taking forever on a 1000 x 1000 wave front.  So far more than 15 minutes.

gr5:
I think the fix isn't so hard.  We just need to buffer it ourselves and send it out through some other method like fprintf().  It's a bit of a mystery why it is so slow as I think Dale you are already using a buffered stream.  But that is the cost of using other people's libraries - sometimes it "bytes" you.

In windows, the file is created almost immediately and you can right click and look at properties and watch the file size grow slowly over many minutes.

gr5:
Fixed.  Will be in next release > 7.2.1

PR #157, opened 2024-08-01
fixes issue #132 (saving wavefront on network drive is slow)

gr5:
It turns out std::endl is also a command to flush out buffers!  Didn't know that.

A 300X300 wavefront was taking 51 seconds to save on my network drive.  After my simple change (check it out!  Such a simple change!) it takes < 1 second.  Not sure how long as it's too quick to measure very well.

I did a binary diff on the wavefront files created old way versus new way and they are identical.

I don't know how this affects unix wavefront files - I believe currently on the PC it stores CRLF in the wavefront file but in linux, only LF.  I don't think my change will affect this behavior.

### DIFF
--- a/surfacemanager.cpp
+++ b/surfacemanager.cpp
@@ -918,9 +918,9 @@ void SurfaceManager::writeWavefront(QString fname, wavefront *wf, bool saveNulle
         for (int row = wf->data.rows - 1; row >=0; --row){
             for (int col = 0; col < wf->data.cols ; ++col){
                 if (saveNulled)
-                    file << wf->workData(row,col) << std::endl;
+                    file << wf->workData(row,col) << '\n';
                 else {
-                    file << wf->data(row,col) << std::endl;
+                    file << wf->data(row,col) << '\n';
 
                 }
             }


gr5:
Adding \n just adds one or two more characters to the buffer.  std::endl has special meaning and it fully writes the buffers out to the disk and waits for that to complete.  So it is writing just a dozen or so bytes into a buffer before flushing the buffers out to the hard drive.  When you remove endl then it continues to buffer until the buffer is full or you close the stream.  Buffers are typically 256 or 1024 bytes so about 20x to 100X faster.

atsju:
According to cppreference, your fix is very clean.
Simply ensure the final flush is done correctly. This is the case if the file is closed properly.

https://en.cppreference.com/w/cpp/io/manip/endl

githubdoe:
Now that all makes sense and brings back an ancient memory.  I think I read that once as well but it did not stick in my mind when the issue came up.

gr5:
I was trying to create a buffer by putting the information in a std::string but I couldn't add the std::endl to the string.  So I read the documentation on std::endl and that's when I realized it was flushing so I started over and just changed those to \n.

PR #184, opened 2025-05-14
A ton of changes by Dale.  Lots of new features.  See VersionHistory for version 7.3.0

gr5:


DFTFringe_Dale.pro
```diff
@@ -30,9 +30,11 @@ SOURCES += main.cpp \
     annulushelpdlg.cpp \
     arbitrarywavefronthelp.cpp \
     arbitrarywavwidget.cpp \
+    astigpolargraph.cpp \
     cpoint.cpp \
```
atsju, line 34: OK I had a quick look during meeting; The windows build issue is linked to AQT and this needs to be managed by me.
However the linux build fails for good/code reason.

Remember we kept "DFTFringe_Dale.pro" to not bother Dale too much with versions and path updates "it works on my machine" type of things.
However the reference file from build point of view is "DFTFringe.pro". If need sources have been aded to "DFTFringe_Dale.pro", same changes need to be applied to "DFTFringe.pro". Please do the updates to fix linux builds.

gr5: I'll take care of that and push it back to dalework.

gr5: If the linux build failed I would have fixed that myself and not bothered you but it didn't try to run that one.

gr5: Fixed, I hope.  I can't easily test it so we'll see if it's fixed when the automated build is working again.

atsju:
If you have some time, can you look at the newly introduced build warnings ? : https://github.com/githubdoe/DFTFringe/actions/runs/15024125124

They are not particularly bad, but they may mask really bad warnings so they need to be gone.

gr5:
> They are not particularly bad, but they may mask really bad warnings so they need to be gone.

I looked at them all.  Well what I mean is I compiled it myself in the qtcreator ide and got 29 warnings.  I looked at all 29.  There may be one that you linked to that isn't showing up with compiler settings I'm using.

Anyway nothing serious. I could fix most of them quite quickly I admit.  I don't think I want to at this point.  When I'm building something and I see warnings, I fix them but I only notice them if they are happening on the files I'm editing (because only those ones get built).

And Dale seems to be ignoring them anyway.  So I don't plan to fix them at this time.

githubdoe (approved):

main.cpp
```diff
@@ -101,7 +102,7 @@ int main(int argc, char *argv[])
     spdlog::flush_every(std::chrono::seconds(3));
 
     // Set the logging format
-    spdlog::get("logger")->set_pattern("[%Y-%m-%d %H:%M:%S.%e] [%^%l%$] %v");
+    //spdlog::get("logger")->set_pattern("[%Y-%m-%d %H:%M:%S.%e] [%^%l%$] %v");
```
atsju, line 105: this must be uncommented in release

astigpolargraph.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+astigPolargraph::astigPolargraph(    QList<int>list, QWidget *parent) :
+    QDialog(parent),
+    ui(new Ui::astigPolargraph), m_list(list)
+{
+    ui->setupUi(this);
+    //For some reason the original starter code of makeChart was in this function but it caused a crash.
+    // I could never figure out why because that code was taken from a working Qt example.  When
+    // moved to inside another function it worked.  So there you go.  Exact same code with no changes worked there but not here.
+    makeChart();
```
atsju, line 93: this makes me super uncomfortable. Might just break later on if we don't understand it.

atsju (commented):
the PR is too long to be review-able.
I will to look at the 3 comments I mentioned. Please delay the release just a little bit.

main.cpp
githubdoe: Sorry it was part of my attempt to disable that for my debugging.  I don't remember but maybe the ifndef technique I used for part of that did not work there for reasons I did not understand nor could get to work.  IIRC

astigpolargraph.cpp
atsju: I tested on linux and got no issue. I will try the linux build also. So far I did NOT reproduce your issue.

main.cpp
gr5: I had noticed this but when I looked at the log output there was date and time included.  Maybe I should have looked in the log file?  I didn't think to.

gr5:
I spent an hour testing.  No problems so far.  Want to test a few more things.

gr5:
I don't see any log file at all :cry:

PR #185, opened 2025-05-14
fix build process. force use python 3.13 for aqt install

atsju:
for reference, I opened an issue in aqtinstall repo. https://github.com/miurahr/aqtinstall/issues/917

Problem is probably linked to lack of support of python 3.9 in lastest AQT package version. But python 3.9 goes end of life soon anyway
https://devguide.python.org/versions/

for more insights on how python is cached in runners : https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-python


gr5 (approved):
Thanks for this.

PR #188, opened 2025-05-16
Changed to 7.3.1 because 7.3.0 will not be released.

gr5:
I had to change the version number to 7.3.1 because I can't untag and then retag the latest master (at least I don't think so).  So the release will be named 7.3.1.  Not certain we are releasing as is.  There may be one more code change.

### DIFF
--- a/RevisionHistory.html
+++ b/RevisionHistory.html
@@ -419,7 +419,7 @@ Displays error of pixel under contour plot cursor and move the profie angle to m
 </ul>
 </ul>
 
-<ul><li>Version 7.3.0</li>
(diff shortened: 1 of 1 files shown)


githubdoe (approved):

atsju:
It's possible to change tags on remote but not recommended.

If you want to create draft releases for test purpose, you can just do tags like v1.2.3-alpha1 or whatever respects server.org   
This way you don't have to change release notes. Once ready redo a tag v1.2.3 on same commit and publish actual release. There will be 2 tags on same place (plus all the alphas123/betas...) but that's something usual

PR #190, opened 2025-05-17
fix remaining build warnings

atsju:
This is probably fine. You can release without the changes if you don't want to make a new tag.
I make the PR as draft and will double check with Valgrind.

### DIFF
--- a/DFTFringe.pro
+++ b/DFTFringe.pro
@@ -18,7 +18,7 @@ QT += charts concurrent core datavisualization gui multimedia multimediawidgets 
 
 qtHaveModule(printsupport): QT += printsupport
 
-QMAKE_CXXFLAGS += -std=c++11
+QMAKE_CXXFLAGS += -std=c++14
 
 # disable qDebug() in release
 CONFIG( release, debug|release ) {
--- a/astigstatsdlg.cpp
+++ b/astigstatsdlg.cpp
@@ -187,7 +187,7 @@ class RunningStat
     {
     public:
         RunningStat() : 
-            m_n(0), m_oldM(NAN), m_oldS(0.), m_min(DBL_MAX), m_max(DBL_MIN) 
+            m_n(0), m_oldM(NAN), m_oldS(0.), m_newS(0.), m_min(DBL_MAX), m_max(DBL_MIN) 
             {}
 
         void Push(double x)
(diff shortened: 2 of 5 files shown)


atsju:
OK I have finally been able to run Valgrind. There is a leak but not here.

there: 
https://github.com/githubdoe/DFTFringe/blob/90c4cbde43546b1f5d52d30a064f0a0b6cb153e6/mainwindow.cpp#L2077
No parent is passed, so this never gets deleted.

Also, during my test I saw that you cannot close DFTFringe as long as the astigPolarGraph window is open. In fact you can not do anything in DFTFringe with this window open. Is this expected behavior ?

DFTFringe.pro
```diff
@@ -18,7 +18,7 @@ QT += charts concurrent core datavisualization gui multimedia multimediawidgets
 
 qtHaveModule(printsupport): QT += printsupport
 
-QMAKE_CXXFLAGS += -std=c++11
+QMAKE_CXXFLAGS += -std=c++14
```
gr5, line 21: omg!  I tried to do this change a year ago to fix a bug releated to unicode characters in file names.  I struggled with this and I couldn't get dftfringe to complile.  QT did not like this change so I had to change it back. And you just change one character and it compiles fine!  Okay this is one thing that sucks about qt creator.

This should fix the unicode character bug in file names issue as Dale's code uses stdlib file functions (fileInputStream or something like that).

Unfortunately I can't do this same change in qt creator unless there is a newer version of it that supports this.

I'm going to start the release process before this merge because I'm worried about what else might break when changing to a newer compiler and I wasn't planning on doing another multi-hour QA session.

gr5:
I tagged master and started the build process so this can be merged now if you want.  I reviewed the changes so far and I approve but I didn't click the review button as I'm not sure if you have more warnings to fix.  Anyway there is no rush as these won't make today's release.

atsju:
about C++14 this is linked to compiler. If you download the compiler with aqt and link it into any old QTcreator it should work. I build with QTcreator on windows too.

Maybe Dale could update his build setting to use DFTFringe.pro instead DFTFringeDale.pro ? I know you have your build setup on your computer but now that I have the whole process explained in readme you should be able to switch. It would limit discrepancies between local and remote build.

gr5 (approved):

PR #194, opened 2025-05-31
Made test stand astig removal window a little taller

gr5:


githubdoe (commented):
No.  that is not the correct fix.  
The window is resized by software using the screen size.  see line 49 of  standastigwizard.cpp.  I recently added that because sizes based on pure pixel width create windows too small for my machine.  I think if I had time I would go through the whole code and change everywhere the UI set width and height for dialogs. and replace them with code like line 49.

githubdoe (approved):
OK,  I guess what you did would have been fine.  sorry.

PR #198, opened 2025-06-12
Dalework

githubdoe:
Updated percent correction feature.

### DIFF
--- a/RevisionHistory.html
+++ b/RevisionHistory.html
@@ -432,11 +432,18 @@ Displays error of pixel under contour plot cursor and move the profie angle to m
 <li>Remembered last ROC offset value in Ronchi and Foucault feature to remember last setting when wave front is changed to a different value</li>
 </ul>
 </ul>
-d
+
 <ul><li>Version 7.3.2</li>
 <ul>
 <li>Added hover and click to astig polar plot to select the wave front and show it as the currently selected wave front.  Added click to astig polar table to highlight the line on the plot and display as current wave front.</li>
 <li>Corrected test stand astig removal feature to display the correct image sizes on the report no matter what the screen resolution is.</li>
 <li>Corrected bug when subtracted wave front result is saved and then loaded so that it does not lose the fact to use or not use the null.</LI>
 <li>Updated the subtract dialog to display a reasonable size based on screen resolution.  Improved help information.</li>
+</ul>
+</ul>
+
+<ul><li>Version 7.3.4</li>
+<ul>
+<li>Modified percent correction to use user settable middle exclude radius and added ability to show values in inches.</li>
+
 </ul>
--- a/percentcorrectiondlg.cpp
+++ b/percentcorrectiondlg.cpp
@@ -34,10 +34,14 @@ percentCorrectionDlg::percentCorrectionDlg( QWidget *parent) :
     QSettings set;
     ui->minvalue->blockSignals(true);
     ui->maxvalue->blockSignals(true);
+    ui->exclusionRadius->blockSignals(true);
     ui->minvalue->setValue(set.value("percent_correction_min",-10).toDouble());
     ui->maxvalue->setValue(set.value("percent_correction_max", 120).toDouble());
+    m_exclusionRadius = set.value("correction exclusion", 19).toDouble();
+    ui->exclusionRadius->setValue(m_exclusionRadius);
     ui->minvalue->blockSignals(false);
     ui->maxvalue->blockSignals(false);
+    ui->exclusionRadius->blockSignals(false);
     ui->maxOrder->blockSignals(true);
     m_maxOrder = set.value("percentMaxOrder",18).toUInt();
     ui->maxOrder->setValue(m_maxOrder);
@@ -50,7 +54,7 @@ percentCorrectionDlg::percentCorrectionDlg( QWidget *parent) :
     sizes << 500 << 100;
     ui->splitter->setSizes(sizes);
     m_number_of_zones = set.value("percent number of zones", 5).toInt();
-
+    m_exclusionRadius = ui->exclusionRadius->value();
     ui->numberOfZones->blockSignals(true);
     ui->numberOfZones->setValue(m_number_of_zones);
     ui->numberOfZones->blockSignals(false);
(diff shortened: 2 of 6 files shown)


gr5:
You didn't pull from master so there was a small conflict.  I fixed the conflict in dalework but now the repository version of dalework doesn't match yours.  Please grab dalework asap otherwise you might start to do more edits to dalework and then there will be more conflicts:

git pull

Also after this pull request is approved and merged, there is a very small edit that you didn't get.  So after this PR is approved do:

git checkout master
git pull                        (this updates your copy of master)
git checkout dalework
git merge master  (merges in the tiny edit)

githubdoe:
I intended to also save and reload the exclusion radius from the zone file.  But don't remember if I completed that.  Will check.

gr5:
Here is gitk -all

It's not obvious but if you mouse over where I put the dashed red line, the cursor changes and you can click and drag to make the top half larger.  I think it's worth understanding this tool as it will help you understand branches better.

Julien mentioned "git fetch" which I don't do often but have many times in the past.  It grabs all the updated branches from github to your machine so you can see both your own branches and the "tracking branches" which match what's on github.  "git pull" does a fetch for only your current branch name and also moves your current branch to the latest commit.  "git fetch" doesn't change anything you might be working on - it just updates your repository's tracking branches.  Not the working branches.  So it's super safe.  Sometimes I do it before a git pull because then I can see if anyone has updated some branch before I do a pull or merge.

image

githubdoe:
"Percentage"  is my short cut way of talking about the "percentage correction" feature,  then when we are currently discussing.

The registry saves the last settings of this feature.  One does not need to save those setting in a file if they do not want to.  That probably is the usual case.  They can change them and the registry will always have the last values.  However maybe they want to recall settings they used sometime ago on a different mirror that the saved in a file.  They can do that provided they saved them.  Which is what the load and save is all about.  Just like the load and save of the mirror config.  It too remembers your last used values even if you don't save them into a file.

The current mirror value has the values of the mirror that must be used by the percent correction feature so I decided to let the profile plot feature always give those to the percent correction feature when it wants a percent correction graph.  That way the current mirror is always what is used.  Those parameters in the zone file a moot now and not used.  I will look at removing them in the save.

githubdoe:
Oh, yes no I remember saving those in the zone file and the registry gives me a way to know if the mirror parameters are still the same as before.  If not the feature should ask if you need new zones.  That last part never got reimplemented.

githubdoe:
There was a race condition that allowed the user to save zones before knowing the ROC.  I don't think that can happen now.  So George perhaps your zone file was created too long ago.  I think recent version do not have that issue.  At least my test of it saved the current mirrors parameters.  As I said earlier I will use them to notify the user that the mirror has changed and see if they want to update the last zones used.

gr5:
Okay I under stand what you are saying now and I now understand what you were saying before as well.  Let me know when you've pushed the changes to dalework (don't forget to make your edits in dalework instead of in master).

Tomorrow I have more time than usual to look at your changes.

I think the ROC got set to 0 in an older version in regedit and the changes persisted.  So anyone who has the older version 7.3.3 will probably have an ROC set to 0 unless your new code notices it's a new mirror and prompts the user.

githubdoe:
Here is the git status which has a problem with revisionHistory.html.  Why the problem?  Why does it say "Both modified" ?  I thought I had done a merge with the remote and so expected it to merge.  It had a merge conflict and I manually edited that conflict (I thought).  I saved that edit I think.  Then this error.

gitstatus

githubdoe:
I don't understand why what I did before to do a manual merge did not work.  It did not work the second time either but has on the 3rd try.
So I pushed my changes.

gr5:
After you fixed the merge conflict maybe you have to create a commit?  I don't remember.

gr5:
I can see that master has been merged into dalework so I think you are in good shape and can make more edits now as needed.

I think the most recent merge conflict maybe had to do with blank lines being different in master versus dalework.  I'm not certain.

githubdoe:
I tried gitK and got it to enlarge the display.  Yet it is just a graph of stuff with little meaning to me.  sorry.  It sure did not help solve the issue.  The first problem was my lack of understanding of what an "add" does.  I thought it added a file to the local git archive.   So I was puzzled by why it recommended me to do an add of that file.  It was already in the archive.  Eventually I gave up and did the add.  Then went and looked what an "add" does.  I don't use it directly so that is why I did not know about it.  I always use a commit -am.  I have never really understood what staging does.  Seems like a commit would be enough.  Oh well that's git.  

So really about all I had to do was add then merge, manual reconcile, then add then commit.  Or something similar that eventually cleared up the problem.

I'm done with changes and ready for more testing by others.

gr5:
Right.  Add and then commit.  Well it's easy to just wildly do a "git add *" But then when you do a git status you see you've staged lots of files you don't want to commit (possibly).  So personally I like that it's two steps as you can easily remove files you added accidentally.  Then I do the commit.

I usually have a few files hanging around that I don't want to be in the repository.

I think it's really worth it to understand gitk -all

If you don't understand the rows in the upper window then you are missing some understanding of rows and branches.  And seeing what is bold is very helpful as it shows you what commit you currently have in your working folder.  It shows if you have files added but not committed or worked on but not added.

And it shows you how everything is merged - what was *not* merged is most important.  You can quickly see that "oh - I thought I merged master into dalework but I can see now that the version of master on github didn't get merged in".  Or whatever it is that I may have messed up personally.

When git says you can't push or can't pull, I find the doing gitk -all helps explain the issue.  And I can click on each row and see what exactly the changes were (often just one line of code or even just one character) and see how it's no big deal to just merge everything into my current work branch.

I used to do gitk -all and then after every git command I would go back to that window and refresh it to see what my git command just did.

There are several other free gui's created by other people that show a display similar to "gitk -all".  You could install one of them.  By the way, have you noticed when you are editing a file in QT it sometimes shows the person, and the date of when that line of code was edited?  It's getting that from git.  In gray to the right of the line of code.

I'm traveling today and "celebrating father's day" but I may have time this morning or much later today.

gr5:
When you say it's all set you mean the update you did last night at 6:15 to  6:19pm, right?  Lots of changes committed then.  Nothing since.

githubdoe:
Yes,

gr5 (approved):

PR #205, opened 2025-07-16
Improve crashlog with minidump

### LINKED ISSUE #203: Improve crashlog
#196 for reference.

All crash need to contain enough information to debug. In 7.3.4 opening wave front inspector without wavefront loaded generates crash without any information. Probably because we catch unhandled exceptions but not segfaults and other. Need to be enhanced.

Discussion with chatGPT told me we should use SetUnhandledExceptionFilter() in additon to std::set_terminate for windows machine. And the we could MiniDumpWriteDump to have a minidump in addition to stack trace.  
To be tested.

atsju:
Close #203 

This adds a minidump to all crashes.   
Crashes from unhandled exception catched with `std::set_terminate` will have stacktrace and minidump.  
Harsher crashes catched with `SetUnhandledExceptionFilter` with only show minidump as spdlog doesn't work during these crashes. To reproduce : open DFTFringe, load no wave front and click on tools -> wave front inspector

- [ ] use `SetUnhandledExceptionFilter` to catch crashes that are not handled today
- [ ] document how to use the minidump to exploit information
- [ ] try exporting registry keys in case of crash

This PR should help us debug unexplained crashes.

### DIFF
--- a/DFTFringe.pro
+++ b/DFTFringe.pro
@@ -55,6 +55,7 @@ win32 {
     LIBS += -L..\build_openCV\install\x64\mingw\bin -llibopencv_highgui460
     LIBS += -L..\build_openCV\install\x64\mingw\bin -llibopencv_imgcodecs460
     LIBS += -L..\build_openCV\install\x64\mingw\bin -llibopencv_imgproc460
+    LIBS += -ldbghelp # for SetUnhandledExceptionFilter
 
 
     # This is for armadillo to not use wrapper. See https://gitlab.com/conradsnicta/armadillo-code#6-linux-and-macos-compiling-and-linking
--- a/main.cpp
+++ b/main.cpp
@@ -26,10 +26,63 @@
 #include "spdlog/sinks/stdout_color_sinks.h"
 #include "boost/stacktrace.hpp"
 
+#ifdef _WIN32
+    #include <Windows.h>
+    #include <DbgHelp.h>
+    #include <ctime>
+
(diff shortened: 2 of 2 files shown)


gr5:
I plan to make sure this builds fine the way Dale and I build. today or tomorrow.

atsju:
UPDATE: I tried to analyse my minidump with windbg tool. It gives some information about the exception but because we build with mingw and not MSVC, the .debug file cannot be used by windbg to fully unveil it's potential. However the minidump contains a stacktrace that we could possibly analyse with addr2line (for the "wave front inspector" crash there was no meaningful info so I'm really not sure)

I see 2 path of action:
- integrate this PR and keep minidump as it doesn't harm. 
- remove the minidump thing. But keep `SetUnhandledExceptionFilter` to at least try to log something in a text file and not have silent crash (spdlog doesn't work in this context)

I would lean for the first. What do you think ?

gr5:
I don't have an opinion so I'll trust you to choose.

atsju:
> @atsju did you see my comment above?

I did. I did not decide. 
On one side it doesn't harm. On the other side, it's one more rarely used (and thus untested) feature.

Let's keep it and see if it brings something on futur crashes

gr5:
Maybe I wasn't clear in the comments above.  This won't compile for me because you added something to the "normal" .pro file but forgot to add it to the dale .pro file.  That's why I haven't approved this PR yet.

gr5 (approved):

gr5:
Okay so I made the same mistake dale did.  I did an inline comment among the code changes and wondered why Julien ignored it.  Anyway he took care of it so this is ready to merge.

Issue #196, opened 2025-06-09
Crash after starting compute


KineticPrecision:
**Describe the bug**
After importing an image (we tried two), and setting up for compute, the program quits after hitting the Compute button.

**To Reproduce**

1. Import image, look at intensity, choose color (green).
2. On the DFT boundary setting, set circle to around center spot.
3. Hit Compute button.
4. Program quits.

**Expected behavior**
Expected a fringe analysis.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 11
 - DFTFringe version: 7.3.3
 - Build: from GitHub release

**Additional context**

Found this in the logs:

```
[2025-06-09 11:55:01.433] [critical] CV error :ranges[i] && ranges[i][0] < ranges[i][1] in D:\a\DFTFringe\DFTFringe\openCV\modules\imgproc\src\histogram.cpp on line 202
[2025-06-09 11:55:01.433] [critical] Unexpected issue. Stacktrace:
```

NOTE:  This computer does NOT have a D: drive.

Found the same issue by installing and running version 6.3.1.

[log.txt](https://github.com/user-attachments/files/20657564/log.txt)

-----

Installed on a different Windows 10 machine, and it ran perfectly.

gr5:
I'll translate the log file and see what I find.

That error that mentions the D: drive has to do with where the file was when it was compiled which should have been on some drive on github somewhere I think?  Not sure but the other lines of the stack trace will probably tell me much more.

atsju:
Hi. Yes D: is GitHub actions build folder. To translate log see here https://github.com/githubdoe/DFTFringe?tab=readme-ov-file#how-to-analyse-crashlog
Usually only few first and few last lines are useful.

I'm sorry to not help more. I have no access to computer

gr5:
Here is the relevant part of the stack trace. In **bold**, I added comments; mostly file and line number gleaned from the addr2line tool.


> [2025-06-09 11:53:55.396] [info] DFTFringe v7.3.3 started
> [2025-06-09 11:53:56.307] [trace] Log level changed to: Info
> [2025-06-09 11:55:01.433] [critical] CV error :ranges[i] && ranges[i][0] < ranges[i][1] in D:\a\DFTFringe\DFTFringe\openCV\modules\imgproc\src\histogram.cpp on line 202
> [2025-06-09 11:55:01.433] [critical] Unexpected issue. Stacktrace:
[...]

PR #207, opened 2025-07-19
Fixed bug in pixelstats.cpp that was causing intermittent crash

gr5:
Fixes issue #196


gr5:
This is the fix that AI recommended and also Dale tested the same change.  @githubdoe or @atsju please look at the files changed tab above and approve this PR.

atsju (approved):
I have read a bit of the openCV documentation and from my understanding, this is indeed programmatically correct AND is what we expect to see displayed.

PR #208, opened 2025-07-20
fix code that mixes wavefront index and name

atsju:
This has been introduced in #193 

I got complain from newer compiler while migrating to QT6 and I think he is right. One more reason to move.
Actually `doThese` contains index of wavefront. And it's compared to wavefront name which makes no sense. 

Please confirm my (AI powered, human checked) proposal is what you intended to do. It would be great to infer the impact of this to add to changelog bugfix.

### DIFF
--- a/surfacemanager.cpp
+++ b/surfacemanager.cpp
@@ -1788,8 +1788,9 @@ void SurfaceManager::subtractWavefronts(){
     QList<QString> list;
     QList<int> doThese =  m_surfaceTools->SelectedWaveFronts();
     for (int i = 0; i < m_wavefronts.size(); ++i){
-        if (!m_wavefronts[i]->name.contains(doThese[0]))
+        if (!doThese.contains(i)) {
              list.append(m_wavefronts[i]->name);
+        }
     }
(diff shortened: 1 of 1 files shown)


gr5:
There was definitely a bug with that line of code.  So you changed the intention of the code a little.  here's what it looks like with your new code below.  I selected the top 2 waves (9814, 9815) and clicked "subtract wave front" which is the only place this code you edited is used.

As you can see it only gives me the option of subtracting out the remaining wavefront(s) which is just wavefront 9816.

The "old" broken code never subtracted out anything from the gui list.  This bug goes back way before #193.  For example, the bug exists in version 7.2.3

But the old code appears to try to only subtract out the first selected wave.

I like the way your new code works.  I just want to double check with @githubdoe that he likes this new method where it removes all selected wavefronts from the gui list.  Do you agree that this is good dale?  Should the gui show:
1) all wavefronts
2) all wavefronts minus the selected wavefronts
3) all wavefronts minus the first selected wavefront

I think (2) is what we want and (2) is what Julien did.




Capture

gr5 (approved):

PR #212, opened 2025-07-24
improve after #211 comments

atsju:
Just slight cleanup after Dales comments made me look deeper into some parts.

Thank you for your review on #211 @githubdoe . Sorry it was a bit complex


gr5 (approved):
To the best of my understanding this is extremely safe and doesn't even warrant me building the project.

atsju:
merging this as it's linked to good questions from Dale in #211

Issue #191, opened 2025-05-23
migration to QT6


atsju:
Hi Dale and George,

As you probably know, Qt5 reached end of life in 2023, and extended support is ending this month. While the source code will remain available, it's likely that the online documentation and related resources will eventually be removed — if not immediately, then probably within the next 24 months.

![Image](https://github.com/user-attachments/assets/b0d83820-e701-4228-bd50-ead524c98ccb)

Do you already have a plan regarding this deprecation?

On my side, I experimented with a migration to QT6.8 LTS on a development branch to see what it would take. It looks like I was able to complete the migration successfully.

You can try a working installer here:
👉 [Download the installer](https://github.com/atsju/DFTFringe/actions/runs/15204446750/artifacts/3183138911)

Due to dependency constraints, I had to update the compiler, Qt, Qwt, and OpenCV together to make everything work properly.

I plan to open a pull request once I've cleaned up the code and updated the documentation.

What are your thoughts on this?

Dale – to ease a potential transition on your side, I can provide detailed documentation on how to recreate your build environment with the updated dependencies (including OpenCV, where your current version differs from the one used in our distribution).

Alternatively, I can send you a pre-configured folder with everything set up. That way, you could simply open the DFTFringe.pro in Qt Creator, only needing to link the updated Qt and compiler versions.

Looking forward to your feedback.

atsju:
Hi @githubdoe, any feedback on this ? I noticed QT5 documentation is already gone from QT website. I'm relieved I made the porting with the documentation still easily accessible.

gr5:
I was thinking it wouldn't be hard to write a bash script or a c++ program to automatically convert the entire stack trace.  However it would be quite slow.  addr2line (or whatever the tool is called) can take sometimes 5 seconds per conversion and a stack trace can typically have 20 lines to be translated.

But it also occurred to me that the program itself could do the conversion as part of the log creation.  But having it sit there for a minute while it's crashing would be very annoying.

atsju:
> But it also occurred to me that the program itself could do the conversion as part of the log creation. But having it sit there for a minute while it's crashing would be very annoying.

Plus the risk of the process failing. The logging shall be extra safe, thus simple.

atsju:
> Note to myself : this might solve the GCC and C++23 version deadlock that made me use boost library instead https://en.cppreference.com/w/cpp/utility/basic_stacktrace.html
> 
[...]

PR #213, opened 2025-07-27
update SPDlog library

atsju:
linked to #191 
separate from #214 for better readability (we do not review this, we just check it still logs)

No impact on DFTFringe code. This is only the logging library code


gr5 (approved):

atsju:
I will merge this as it will not impact Dale in any way. 
There shall never be merge conflict here even if you do not pull the branch locally.

PR #214, opened 2025-07-27
Minor Changes necessary for QT6 that are compatible with QT5

atsju:
As mentioned in #191 QT5 support is slowly falling appart.

However I don't want to break development flow for Dale. So here is a set of changes that are compatible with QT5.
I want to integrate them now, so the futur move to QT6 is easier and I can maintain the QT6 branch on my side until we decide to move.

@gr5 there is one change related to OpenCV that I only tested with CV 4.6.0. However Dale is building with CV 3.4.12. Could you test if this compiles as I suppose you already have this version available ?

### DIFF
--- a/camwizardpage1.cpp
+++ b/camwizardpage1.cpp
@@ -240,11 +240,11 @@ void CamWizardPage1::on_compute_clicked()
                 if (found)
                     break;
                 if ( i == 1){
-                    cv::SimpleBlobDetector detector;
+                    static Ptr<SimpleBlobDetector> detector = cv::SimpleBlobDetector::create();
 
                     // Detect blobs.
                     std::vector<cv::KeyPoint> keypoints;
-                    detector.detect( dial, keypoints);
+                    detector->detect( dial, keypoints);
 
                     /*int horz = 1;
                     int vert = 1;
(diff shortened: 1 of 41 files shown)


camwizardpage1.cpp
```diff
@@ -240,11 +240,11 @@ void CamWizardPage1::on_compute_clicked()
                 if (found)
                     break;
                 if ( i == 1){
-                    cv::SimpleBlobDetector detector;
+                    static Ptr<SimpleBlobDetector> detector = cv::SimpleBlobDetector::create();
 
                     // Detect blobs.
                     std::vector<cv::KeyPoint> keypoints;
-                    detector.detect( dial, keypoints);
+                    detector->detect( dial, keypoints);
```
atsju, line 247: this is where it might eventually fail with older CV version.

gr5: When I try to test this code, I put in a breakpoint and used the circle grid and it finds it on line 240 above so it rarely gets to lines 243,247 which you may be worried about.  It did get to line 247 once if I search for too many circles but it still seems to find them all. 

 I'm not sure what might go wrong?  Compiler error?  There was no problem compiling the code.  I think it's working just as well as it used to.

contourrulerparams.cpp
```diff
@@ -35,7 +35,7 @@ void ContourRulerParams::on_radialAngleSB_valueChanged(int arg1)
 void ContourRulerParams::on_setColorPB_clicked()
 {
 
-    QColor color = QColorDialog::getColor( ui->setColorPB->palette().color(QPalette::Background));
```
gr5, line 38: I never used the contour ruler before.  This color picker works fine and it does indeed set the contour ruler color.

camwizardpage1.cpp
atsju: If anything I expected a compilation error. So this is probably fine.

I double checked and I did this because simpleblobdetector became an abstract class in 4.7.0 so this is really fine with older versions too.

This also means I can bump CV version in automated build with these changes. It won't clash with Dale project file.

generatetargetdlg.cpp
```diff
@@ -38,17 +40,9 @@ void generateTargetDlg::on_generate_clicked()
...
-
-       vsize = qApp->desktop()->height();
-       vsize /=2;
-       vsize *= 2;
-
-
-
+       QScreen *screen = QGuiApplication::primaryScreen();
+       hsize = screen->geometry().width();
+       vsize = screen->geometry().height();
```
atsju, line 45:
```suggestion
       // make sure sizes are even
       QScreen *screen = QGuiApplication::primaryScreen();
       hsize = screen->geometry().width();
       hsize /= 2;
       hsize *= 2;
       vsize = screen->geometry().height();
       vsize /= 2;
       vsize *= 2;
```

gr5: Go ahead and commit it and I'll rebuild just to be extra safe, lol.

PR #215, opened 2025-08-02
prepare release 7.4.0

atsju:
remove a thing introduced in #201 that shall not appear to end user.
IMHO This new auto invert option should be in preferences. Not in the main menu.

Anyway: new feature measure minor version update not only patch. => v7.4.0

### DIFF
--- a/RevisionHistory.html
+++ b/RevisionHistory.html
@@ -446,5 +446,19 @@ Displays error of pixel under contour plot cursor and move the profie angle to m
 <ul><li>Version 7.3.4</li>
 <ul>
 <li>Modified percent correction to use user settable middle exclude radius and added ability to show values in inches.</li>
+</ul>
+</ul>
+
+<ul><li>Version 7.4.0</li>
+<ul>
+<li>New button to remove wavefront auto invert.</li>
+<li>Fixed bug that was causing intermitent crashes.</li>
+<li>Fix discrepency in wavefront 2D cut angle versus wavefront profile.</li>
+<li>Fix wavefront subtraction.</li>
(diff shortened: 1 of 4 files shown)


RevisionHistory.html
```diff
@@ -446,5 +446,19 @@
 <ul><li>Version 7.3.4</li>
 <ul>
 <li>Modified percent correction to use user settable middle exclude radius and added ability to show values in inches.</li>
+</ul>
+</ul>
+
+<ul><li>Version 7.4.0</li>
+<ul>
+<li>New button to remove wavefront auto invert.</li>
```
gr5, line 454:
```suggestion
<li>New menu item to reset wavefront auto invert.</li>
```

RevisionHistory.html
```diff
@@ -446,5 +446,19 @@
...
+<li>Fixed bug that was causing intermitent crashes.</li>
+<li>Fix discrepency in wavefront 2D cut angle versus wavefront profile.</li>
+<li>Fix wavefront subtraction.</li>
```
gr5, line 457:
```suggestion
<li>Improved wavefront subtraction.</li>
```

gr5, line 457: Looking at the code it was a bug but for the users, it's just an improvement where before you could subtract any of the wavefronts but now you can't subtract a wavefront from itself.  So I'm calling it an "improvement".

gr5 (approved):
looks great.  You can tag it and make the release but please don't finalize it until I've tested the installer and the released version.

gr5:
Okay then, well if you merge this I will push the tag.

PR #216, opened 2025-08-06
QT6

### LINKED ISSUE #191: migration to QT6

atsju:
Here we go. Fully compatible with Qt5 and Qt6. CI/CD with installer generation for both.
Next release should be named v8.0.0 (no hurry on that).

Build on:
- QT6.8 (LTS) instead QT5
- QWT 6.3.0 instead 6.1.6
- OpenCV4.12.0 instead 4.6.0

Using:
- C++17 instead C++11
- mingw13.1.0 instead 8.1.0

Close #191


githubdoe (approved):
Nice work.  Thank you.

PR #226, opened 2025-08-14
fix reading mirror config ellipse

### LINKED ISSUE #221: mirror config file doesn't save elliptical state
If you check the ellipse checkbox in mirror config, save the config and load the config, the elliptical checkbox is unchecked.  This bug is in 7.4.0.  The information is stored in the json file but something goes wrong when read back.  I may have introduced this bug.  Not sure.  From json file:

image

atsju:
close #221

I did not check other spellings. 
However this confirms having abstraction classes for this type of things makes sense.


atsju:
I don't know if it's case sensitive. If you look closely you see S and P are inverted also.

Issue #33, opened 2023-06-27
Licences


atsju:
This issue is to track me research.
- QT: https://www.qt.io/download-open-source https://www.qt.io/licensing/open-source-lgpl-obligations open source QT licence is LGPLv3  and GPLv3. 
   - [ ] licence is REQUIRED to be prompted and acknowledged by end user of application => to be confirmed as https://www.gnu.org/licenses/gpl-faq.html#ClickThrough explicitly states it's NOT REQUIRED
   - [x] We also need to provide a way to re-link application (this is OK, the code is available on GitHub). 
   - [ ] And we need to provide a copy of QT sources with the code (or a written way to get the code).
- OpenCV: OpenCV 4.4.0 and lower use "3-clause BSD License" https://opencv.org/license/
   - [ ]  need to share a copy of the licence
- QWT:
   - [ ]  
- Armadillo:
   - [ ]  
- Lapack:
   - [ ]  
- Blas:
   - [ ]

PR #227, opened 2025-08-14
move zapm.cpp to dedicated folder as it's not our code

atsju:
https://github.com/mlpeck/zernike

working on #33


atsju:
@githubdoe can you have a look at this small pull request ? 
All I did is moved zapm.cpp around to keep track it's an external file and add licence file.

---------- 
Additional question:
Out of curiosity I checked if there had been some update on Michael side and it looks like there are small changes from original file.
image
I think you did that on purpose as I didn't find any historical file from Michael matching this. Can you please confirm if this is intentional ?
Maybe I could add some comments about it "Dale modified for <purpose>" so that anybody looking into this in 10 years from now doesn't get too much headache :)

githubdoe:
You can delete it.  I don't need to see it anymore.

githubdoe:
I don't remember those changes.  Sorry..

atsju:
Oh I see now! You copied zapm to zernikeprocess.cpp. thank you

> If I understand perhaps Mike made changes that I did not pick up. I'm not sure which file is mine and is Mike's in you example.
no I checked and that's not the case. https://github.com/mlpeck/zernike/blob/master/src/zapm.cpp I looked at history and Mikes version is like this since the beginning. 

Let me close this pull request. I will look at zernike file deletion. I might put zapm in a separate file just for licence tracking purpose.

atsju:
green is mike. Red is yours

atsju:
Thank you this actually makes more sense now. I did not delete zapm.cpp as proposed by @githubdoe as it's actually used. 
What I did instead is remove the zapmC (old name of Mike zapm) from DFTFringe code as it was not used. I did some cleanup but the code that runs is the same as before.
So this PR grow a little bit because of that but source is same.

What I propose in a different PR to make everything even is to remove zpmC (old name of zpm) and add Mike zpm.cpp as it is done for zapm.cpp with little commented modifications.
This would make it easier to follow I suppose and keep track of authors. I checked code and their are identical so it's more like moving things arround.

gr5 (approved):

githubdoe:
Thank you George for remembering the Normalizing terms.  Yes that is what makes mine different.  The zpm and zpmC were related and I have a memory of the zpmC being reserved for future use but that may be a false memory.  I use to remember what the "C" stood for.  Perhaps that will came back to me.  So now I'm unsure that it should be deleted.  But if I don't remember it's purpose and if George does not as well then it probably does not matter.  Oh I think maybe the "C" stood for circular......But....

atsju:
Dale I checked and it has just been renamed in latest releases. ZapmC became zapm and zpmC became zpm. Those are same. Maybe the C stands just for the language C. As Mike does lot of R this would make sense. Maybe ask him.

What you might remember however was that you implemented both zpm(C) and zapm(C) where "a" probably stands for annular. And you copied the new zapm instead of using your existing zampC (which are the same - remember renaming).

I think this could indeed make sense with your partial memories.

PR #228, opened 2025-08-15
pass by reference or by value when necessary

atsju:
I have used clazy to automatically use refs and values where it should.
`QPoint` is misleading but it's indeed a simple type to pass by value.
Notice there is no big risk changing type a to const type &a and reverse.

Those changes should bring a marginal efficiency gain. Nothing really measurable.

### DIFF
--- a/contourplot.cpp
+++ b/contourplot.cpp
@@ -378,7 +378,7 @@ void ContourPlot::ruler(){
     }
 }
 
-void ContourPlot::selected(const QPointF& pos){
+void ContourPlot::selected(QPointF pos){
     if (m_wf==0)
         return;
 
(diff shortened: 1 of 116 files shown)


contourplot.cpp
```diff
@@ -378,7 +378,7 @@ void ContourPlot::ruler(){
     }
 }
 
-void ContourPlot::selected(const QPointF& pos){
+void ContourPlot::selected(QPointF pos){
```
gr5, line 381: I get that you are adding tons of "const  &" and that makes it so it uses the incoming variable as is but promises not to modify it.

But why are you removing the const & in a few cases such as this one?

githubdoe (approved):

PR #235, opened 2025-08-19
end support of saving .ini file format for mirror configs

### LINKED ISSUE #233: Remove ability to save mirror config in .ini
As discovered in #221 we can still save .ini files for mirror config. (and load them but we decide to keep discussed on #225).

If I understand correctly we should no save in this format anymore. Need to remove the ability to do it and keep saving .JSON. support loading of both.

for reference : json format introduced in #125

atsju:
close #233 

I propose we do the v8.0.0-alpha1 after merging this. 
To gather feedback if Qt6 brings any unexpected change/problem to users.

### DIFF
--- a/dftfringe_fr.ts
+++ b/dftfringe_fr.ts
@@ -2473,7 +2473,7 @@ Wave Length nm</source>
     </message>
     <message>
         <location filename="mirrordlg.cpp" line="335"/>
-        <source>Read mirror configuratoin file</source>
+        <source>Read mirror configuration file</source>
         <translation type="unfinished"></translation>
     </message>
 </context>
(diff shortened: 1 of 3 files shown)


gr5 (approved):
compiled and tested and it works.  I looked at all the source code changes.  That looks good as well.

atsju:
Hi @githubdoe 
I do not see your approval or rejection here (#235) and in #227 #228
Kindly let me know if you need time to review, approve, comment or reject. 
As you answer to other comments I don't know if you missed those PR, if you need time or if something went wrong and we do not see your comments.

Again, no time pressure. Just shoot a little message to let me know what to expect.

PR #236, opened 2025-08-19
update changelog to prepare for 8.0.0-alpha1

### LINKED ISSUE #225: 8.0.0 release. Qt6
This is a placeholder to discuss 8.0.0 release (QT6 based).

I propose to use the major update to do some breaking changes

- ~~#224~~ we decide to keep backCompatibility so this is not required for 8.0.0
- ~~drop support for .ini miror config files. People should use 7.4.x to save their .ini files into .json files. Should not impact too much people anyway~~ Dale wants to keep

Once these changes are done: create a `v8.0.0-alpha1` public release to get feedback if we missed something.
Later we switch to v8.0.0. And if there is anything blocking in the meantime, we can still do a v7.4.x bugfix branch to backport fixes.

Are you OK with my proposal or do you prefer to keep changes minimal ? 
Do you see anything else to add before going to v8.0.0 ?

atsju:
close #225 

First of all: yes I messed with this file. 
I'm sorry but I think it's actually more readable like this. I used an autoformatter and it helped find out many places that had incorrect html tags. html is in fact quite permissive.

So kindly just focus on test of 8.0.0 or look at final result in your browser.

Also I propose that we invert the order of the list. Maybe have 8.0.0 at top of file and v1 at bottom. What do you think ?

### DIFF
--- a/RevisionHistory.html
+++ b/RevisionHistory.html
@@ -1,471 +1,555 @@
-<html><body><h1>DFTFringe Version History</h1>
-<ul>
-<li> Version v1.0  Initial release</li>
-<li> V1.1-beta  corrected saving, inport and export of Mirror dialog with OpenFringe</li> </li>
-<li> v1.2-beta
-<ul>
-<li> Added credits to about</li>
-<li> Reduced size of test stand wizard and made it more resizable</li>
-<li> Save mirror outlines for loaded interferogram every time the "Done" or "make surface" buttons are used.<br>
-Look for saved outline file associated with the loading interferogram and use it if found.</li>
-<li> Internal change to how zernike values are computed.  Now using Polar definitions. </li>
-<li> Increased zernike terms to 48 using all 6th order terms</li>
-<li> Added batch processing of interferograms both manual assist and auto mode.</li>
-<li>Added enable all and enable spherical only buttons to Metrics display</li>
-<li>Added interferogram wavelength to metrics</li>
-<li>Added average of stand/system induced terms to stand astig reporting</li>
-</ul>
-<li> Version 1.3-beta</li>
-<ul><li> Added hot keys for editing the outline.</li>
-<li> Changed outline help to a button to open a docked help window on the left side.</li>
-<li> Corrected and reworked logic to save wavefront stats. Moved display of stats to under the View Menu</li>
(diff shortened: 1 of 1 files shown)


atsju:
@githubdoe if you have time, kindly approve (or reject) the other opened pull requests. No pressure about the timeline I just want to ensure you see them.

githubdoe (commented):
Typo "mirro"  line 550

RevisionHistory.html
```diff
@@ -1,471 +1,555 @@
...
+
+        <li>Version 8.0.0</li>
+        <ul>
+            <li>First release based in QT6 and latest dependencies.</li>
+            <li>Improved crashlog.</li>
+            <li>Slightly faster zernike computation.</li>
+            <li>Distribute DFTFringe with required licences and copyright notices.</li>
+            <li>No more support for saving mirror configuration file in .ini format. Now only .json
+                format is supported for saving. .ini files can still be loaded.</li>
+            <li>Fix wavefront selection in RMS plot. Broken in 7.4.0.</li>
```
gr5, line 549: What's RMS plot again?  Is this where you fixed the spinnable arrow thing that shows different diameters?  Maybe you can explain slightly better.  Also I don't want to scare anyone into thinking all analysis with 7.4.0 were bad and need to be thrown out. So maybe (if it's true) say "fix minor bug...".

The terminology that I think Dale uses:

3d plot (top left plot)
contour plot (top right plot)
profile plot (one (or sometimes 16) diameter of the wavefront(s))

atsju: Spintable was fixe din 7.4.0 : Fix discrepency in wavefront 2D cut angle versus wavefront profile
 
 RMS plot: as Dale explained. Batch process, you see RMS values and "click to select" was broken

gr5 (approved):
I don't care if newest is at the top or the bottom.  Every time I open that file I am surprised that the newest is at the bottom but I don't think users look at it as they see the same info on the releases page on github.

atsju:
> Every time I open that file I am surprised that the newest is at the bottom

I will revert. I agree people probably do not look at it but still...

atsju:
I reversed order of the revisions.

You might want to look at last commit also 312675a. I got Copilot help for review.

atsju:
@gr5 sorry. Additional changes again

gr5 (approved):

PR #237, opened 2025-08-20
remove old style connects

### LINKED ISSUE #229: QT old style connects
I'm using clazy to to update old-style-connect.
More information [here](https://doc.qt.io/qt-6/signalsandslots-syntaxes.html#) and [here](https://wiki.qt.io/New_Signal_Slot_Syntax).
**The new syntax is safer because it's checked at compile time.** But also it should be marginally faster (I measured no difference).
The clazy automated fix seem to work well but there are about 32 other old style connect that I will fix manually.

I think I found several that should not be working in current code. However I don't know how to test them: 
- https://github.com/githubdoe/DFTFringe/blob/db86845d3810657019267ccf971f424a67524500/oglview.cpp#L223
`saveAllContours` is a private slot. Can you let me know how to test this specific line ? I don't know what corresponds to `OGLview` in the GUI.
- https://github.com/githubdoe/DFTFringe/blob/db86845d3810657019267ccf971f424a67524500/simulationsview.cpp#L169 the signature of `saveImage(QString fileName)` doesn't match `QAction::triggered(bool checked = false)`
- https://github.com/githubdoe/DFTFringe/blob/db86845d3810657019267ccf971f424a67524500/rmsplot.cpp#L13 probably broken since #211 (me).
- https://github.com/githubdoe/DFTFringe/blob/db86845d3810657019267ccf971f424a67524500/mainwindow.cpp#L184 `SurfaceManager::backGroundUpdate` is a private slot
- ...
- List will increase when I work on it

atsju:
close #229 

I place this here to show you. Feel free to approve or reject.
I still need to replace one last slot that will require a small rework but is working as is. Hence the "draft" pull request.

Some have been automatically replaced, some manually. You can look at changes commit by commit if you want to dig.
I have found no real issue with original code (except one fixed in #231). This will just make it easier to maintain.

### DIFF
--- a/IgramArea.h
+++ b/IgramArea.h
@@ -169,7 +169,8 @@ public slots:
     void zoomIn();
     void zoomOut();
     void igramOutlineParmsChanged(const outlineParms &parms);
-    void increase( int i = 1);
+    void increaseValue(int);
+    void increase();
     void decrease();
     void zoomFull();
     void toggleHideOutline();
@@ -201,7 +202,6 @@ private:
     QDockWidget *dock;
     void drawBoundary();
     void resizeImage();
-    void createActions();
     void zoom(int del, QPointF zoompt);
     bool modified;
     bool scribbling;
--- a/astigscatterplot.cpp
+++ b/astigscatterplot.cpp
@@ -12,7 +12,7 @@ astigScatterPlot::astigScatterPlot(QWidget *parent):QwtPlot( parent ),m_max(.3)
 {
 
     Picker = new myPlotPicker( this );
-    connect(Picker, SIGNAL(mySelected(QString)),this,SLOT(selectedwave(QString)));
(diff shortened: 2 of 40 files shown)


atsju:
everything that is NOT in first commit e82c3ac has been coded manually. If you want to check everything I recommend you do commit by commit. I tried to make them as reviewable as possible one by one.

> a typo in any of them could cause a bug, right?

Yes but not anything near the way it was before. Now all connect are compile time checked. This means there cannot be "no connection" as before with a warning in the logs. 
Only thing that could happend is that I connected to a wrong signal (same name from different class ?). 
That why I was willing to do the move.

If you think the risk is too high with legacy code I can live with it. Personally I prefer live with an unlikely but possible break in the futur alpha and have a the compile time checks.

atsju:
Last change done. Ready for review

PR #238, opened 2025-08-20
fix forever loading cursor during batch processing

### LINKED ISSUE #21: [bug] loading multiple igram, mouse icon is "loading"
Version 6.2

When loading a batch of igram, I then reduce the window "batch process igrams" (closing aborts the process).
But until I'm finished with batch processing, the mouse is not an arrow anymore but has the circular shape for "loading". It's quite disturbing as the computer is waiting for my input and not for an internal action to finish. When i see the "loading" icon, I generally wait for it to return to normal before doing anything else

atsju:
fix #21 

I think I finally found it. This has been bothering me for so long!

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -1105,7 +1105,7 @@ void MainWindow::batchProcess(QStringList fileList){
     m_contourView->getPlot()->blockSignals(true);
     QSettings settings;
     bool shouldBeep = settings.value("RMSBeep>", true).toBool();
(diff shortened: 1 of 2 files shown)


atsju:
I'm super happy. This has bothered me for over 2 years !
Got a little help from Copilot. It did not find the issue but helped me direct my searches.

atsju:
Hi @githubdoe you might want to have a look at this one before I merge it.

PR #240, opened 2025-08-21
remove qt5core dependency

atsju:
#236 

If we could get completely rid of Qt5 it would be great.

I tested it seems to work as expected


atsju:
ready for review

githubdoe (approved):

Issue #242, opened 2025-08-22
Removing dead code


atsju:
I propose to remove dead code. 

I do not want to remove debug code or things like that. Just code that is here and is really not needed.

atsju:
fixed in #244

PR #243, opened 2025-08-22
enable gc-section for lighter application

atsju:
These flags removes unused functions and variables from final binary. Makes it lighter.
It's safe, I have used those professionally for years.

Wa can still clean the code for readability #242


atsju:
On linux there is a substantial gain but on Windows this results in binary growth.
Probably the additional separate sections make it grow more than what is deleted because unused. 
I suppose it's not worth adding this.

PR #263, opened 2025-11-08
contourview display is now always a perfect 1:1 aspect ratio

### LINKED ISSUE #56: If you zoom in but don't zoom out of right side wavefront window then load new wavefront you can't zoom out
**Describe the bug**
zoom level gets stuck in right wavefront window

**To Reproduce**
Load an igram, process it click in drag in right window to zoom in somewhat
Without right clicking, go back to igram tab and process again (click "done" then "process igram".  now this second wavefront is still zoomed in but it is stuck zoomed in.  Right clicking doesn't zoom out.

**Expected behavior**
Right clicking should zoom back out.  Probably should not start zoomed in.  Maybe.

atsju:
Low priority but nice small change. 
test it out and we can have it in next release

list of issues that still need to be fixed:
- [x] ensure square ratio in basic situations of normal use (load igram, load wft, switch between wfts, rescale window, ...)
- [x] fix #56 noticed by gr5 : if you zoom in on the contour and then load another wavefront, it doesn't zoom out (you are still zoomed in on the same area but on the new wavefront) and if you try to zoom out it won't because it thinks it already is zoomed out.
- [x] change scale of window then zoom and unzoom : results in non-square. New 
- [x] fullscreen. Zoom then unzoom will result in deformed picture.
- [x] acceptable ? zoom doesn't ensure square ratio. zooming rectangle will expand more one direction than other
- [x] acceptable ? when not large enough in X direction, you cannot zoom anymore
- [x] colorscale is correct. Currently uses size as scale
- [x] show all selected wavefronts must work properly
- [x] report generation must work properly

### DIFF
--- a/contourplot.cpp
+++ b/contourplot.cpp
@@ -74,7 +74,7 @@ public:
             return;
         }
 
-        // Get the canvas dimensions and aspect ratio (accounts for axes, labels, colorbar, etc.)
+        // Get the canvas dimensions
         QWidget *canvas = thePlot->canvas();
         double canvasWidth = canvas->width();
         double canvasHeight = canvas->height();
@@ -83,36 +83,24 @@ public:
             QwtPlotZoomer::zoom(rect);
             return;
         }
-        double canvasRatio = canvasWidth / canvasHeight;
 
-        // Get the selected zoom rectangle dimensions and its aspect ratio
+        // Get the selected zoom rectangle dimensions
         double selWidth = rect.width();
         double selHeight = rect.height();
-        double selRatio = selWidth / selHeight;
 
-        QRectF adjustedRect = rect;
+        // Use shared aspect ratio adjustment logic to match canvas aspect ratio
+        QRectF adjustedRect = thePlot->adjustRectToAspectRatio(
+            selWidth, selHeight, canvasWidth, canvasHeight);
 
-        // The zoom rectangle should have the same aspect ratio as the canvas
-        // to ensure no distortion when displayed
-        if (selRatio > canvasRatio)
-        {
-            // Selected rect is wider than canvas → expand height to match canvas ratio
-            double newHeight = selWidth / canvasRatio;
-            double extra = (newHeight - selHeight) / 2.0;
-            adjustedRect.setTop(rect.top() - extra);
-            adjustedRect.setBottom(rect.bottom() + extra);
-        }
-        else if (selRatio < canvasRatio)
-        {
-            // Selected rect is taller than canvas → expand width to match canvas ratio
-            double newWidth = selHeight * canvasRatio;
-            double extra = (newWidth - selWidth) / 2.0;
-            adjustedRect.setLeft(rect.left() - extra);
-            adjustedRect.setRight(rect.right() + extra);
-        }
+        // Translate adjusted dimensions back to rect position
+        QRectF zoomRect(rect.left() + adjustedRect.left(),
+                       rect.top() + adjustedRect.top(),
+                       adjustedRect.width(),
+                       adjustedRect.height());
+
         // Set guard to prevent intermediate aspect-ratio updates while zooming
         thePlot->m_inZoomOperation = true;
-        QwtPlotZoomer::zoom(adjustedRect);
+        QwtPlotZoomer::zoom(zoomRect);
         // Ensure guard is cleared on next event loop tick in case rescale() isn't called
         QTimer::singleShot(0, thePlot, SLOT(clearZoomFlag()));
     }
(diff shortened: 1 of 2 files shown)


atsju:
Dale here you go : https://github.com/githubdoe/DFTFringe/actions/runs/19196205264/artifacts/4508565928

Each time a PR is opened the installer is automatically created. I can make a tutorial for you to access it for you to store and remember or you can ask and I paste link. No problem.

githubdoe:
First attempt gives this display
image

githubdoe:
By the way I actually use the fact the contour displays dimensions values in pixels.  That is the fastest way to check the size that someone used for their wave front analysis.  I would be disappointed if that changed.

atsju:
OK it fails only if you make "files -> read wavefront" first thing. 
It works with igrams and it works if you resize the window.

Probably a small fix. 
image

githubdoe:
Sorry there is text but not the right text.  It appears the scale is y dimensions and not the error height.  Here is the expected.

image

githubdoe:
I used two different small tools on the 16 inch mirror.  Both made on the 3D printer out of PETG.  Made a tape dam around the perimeter and poured the pitch into it.  Both were experiments intended to take down a high zone about 2 inches wide.  It did that but also went a little too far and left several grooves.  

Here is an image of the tools and the contour map under the mirror where I tried to reduce some of the high spots left.  I failed at that.
MAP


Here is the surface before the trials and midway through.

Nov27Sess1

Here is the last attempt.  Large ring is gone and mirror has the lowest RMS ever but many small rings that require smoothing with a doctored large lap avoiding the 75% radius low area.

Low


By the way I have made a change to how the ruler dispalays over teh contour map.  I added a ring every 10 percent.
 I like it a lot compared to the 50% ring I had before.    I will write up a feature request later.  I'm trying out cloning the latest master to see if I can use that for development.  It seems to build ok.  I just have not had anytime to play with it.

Here is the new contour plot with ruler turned on.

image

githubdoe:
Yes that is happening next.  However the full sized tool will be modified as a petal lap and with almost no contact at the 75% zone.  on advice from Mel Bartels who is my mentor on this.

image

gr5:
Looks good in main window but still a problem with this feature here as shown:

image

githubdoe:
Also check the view in the report if you have not already.

gr5:
@atsju - I found a bug.  If I lower the resolution to 500 you can't zoom in anymore.  You did fix my old zoom in bug.  And I like your solution.  But there is a new bug (maybe an old bug?).

So load an igram and when you get to the DFT page, lower the resolution to 500 (I initially found the problem when I lowered it from 600 to 300 but 500 seems to be low enough).  Then click "recompute DFT" button.  Then "compute surface" as you normally would.

Now you can't zoom in on the contourplot.

I think you can get it in a bad state where other contours at 600 stop being able to zoom in but I couldn't duplicate that.

This bug was hard for me to isolate so it can be duplicated every time but here you go - just do what I said above.

I even tried deleting everything in my registry under DFTFringe and restarting dftfringe to be sure this happens every time.

Version 8.2.0 is fine with zooming in when the resolution is 500 pixels.

atsju:
@gr5 I think this is now fixed. It was a race condition in display updates.
Solving it also solver a minor think I had identified "acceptable ? when not large enough in X direction, you cannot zoom anymore"

atsju:
> The other issue is that you seemed to have merged in master so you have my newer code that lets the right tool bar be smaller. But it's now ALWAYS right at the minimum. I can't make it bigger. And when I did that PR, I put the minimum very small so things overlap a bit and you can't quite read some of the buttons and such. It's not ideal.
> 
> When I try to make the right tool bar bigger, the mouse flickers to the other mouse pointer shape and when I try to click and drag, it just does buggy things. But it never allows me to make the right toolbar bigger. Sometimes it does nothing, sometimes it dissappears (and recovers if I just click over there anywhere).
> 
> I'm thinking @atsju that you changed something: maybe the code that deals with the "teal" separators or maybe the code that deals with the mouse dragging to zoom in on the contour plot. Anyway, for some reason I can't change the size of the right tool bar anymore.

I will have a look. I did not change the code in the way you think. However to keep things square during resize, there are many QT events and I now recompute canva and picture ratio when you drag any menu. 
I do always drag the middle separator I did not try the right bar menu. I will try to reproduce and fix.

atsju:
Hi @gr5 

I tried the lastest installer and I reproduce same behavior at 125% I already signaled when reviewing the right tab size PR
https://github.com/githubdoe/DFTFringe/pull/295#issuecomment-3617870576
But I did not reproduce the mouse flickering or anything you described here. I'm able to resize right tab without problem. 

My process is:

- full screen 100%
- load 1 igram
- compute
- move righ side bar
- everything adjusts fine

Could you share a video capture of your test/behavior ? It's not urgent at all though. No pressure!
Also it might be worth it installing master CI build just to confirm the behavior is linked to this PR.

I feel a bit sorry you do all the testing and some things I could have seen. So I really want to thank you for this work and the effort you put here !

gr5 (approved):

PR #272, opened 2025-11-13
Fixes ronchi issues caused by gaussianBlur issues

### LINKED ISSUE #162: gaussian blur causes weird pixelated edges in simulated ronchi
Version 7.2.3 but I think it does it in many older versions as I've seen this before

**To Reproduce**
I processed igrams from a real mirror.  I'll include pictures of the issue and a picture of a ronchigram of the real mirror.

I smoothed the wavefront with "tools" "smooth current wavefront" although that's not required to see the problem.

I created a perfect simulated wavefront and the same issue happens: looks bad if gaussian smoothing is on.

GOOD:

2024-10-09 12_53_47-geo_processed_smoothed

BAD:

2024-10-09 12_53_16-geo_processed_smoothed

ACTUAL:

DSC04479

### LINKED ISSUE #257: ronchi and foucault show too much rough information
Testing with 8.1.0 and I notice for that 100th time.  If I only have one igram and try to look at ronchi and foucault I get lots of fringe "print through".  I can up the gaussian blur to say 25% and even though the 3d view of the wavefront shows zero print through and looks extremely smooth, the simulated foucault looks awful with fringe print through (red circled area below).  Also the outer circumpherence of the foucault is nasty (blue circled area below).

My work around is to always smooth the wavefront (menu tools "smooth current wavefront") and then I also have to disable gaussian smoothing to fix the blue circled issue.

It appears that for the ronchi and foucault, the smoothing is done to the ronchi and foucault images instead of to the wavefront (and then create the ronchi and foucault from the gaussian smoothed wavefront).

**I propose to do the gaussian smoothing first, and then create the ronchi/foucault images instead of the other way around.**

Image

gr5:
This fixes #162 and fixes #257

### DIFF
--- a/simulationsview.cpp
+++ b/simulationsview.cpp
@@ -218,7 +218,7 @@ cv::Mat SimulationsView::nulledSurface(double defocus){
         int blurRad = .01 * gbValue * m_wf->m_outside.m_radius * 2;
         blurRad &= 0xfffffffe;
         ++blurRad;
-        CropGaussianBlur(nulled_surface.clone(), nulled_surface, blurRad, m_wf->m_outside, m_wf->m_inside);
+        CropGaussianBlur(nulled_surface, nulled_surface, blurRad, m_wf->m_outside, m_wf->m_inside);
     }
 
     nulled_surface  *= M2PI * md->lambda/outputLambda;
--- a/surfacemanager.cpp
+++ b/surfacemanager.cpp
@@ -440,7 +440,7 @@ void SurfaceManager::generateSurfacefromWavefront(wavefront * wf){
             //std::chrono::milliseconds ms = std::chrono::duration_cast< std::chrono::milliseconds >(
             //    std::chrono::system_clock::now().time_since_epoch());
 
-            CropGaussianBlur(wf->nulledData.clone(), wf->workData, gaussianRad, wf->m_outside, wf->m_inside);
+            CropGaussianBlur(wf->nulledData, wf->workData, gaussianRad, wf->m_outside, wf->m_inside);
 
             //std::chrono::milliseconds ms2 = std::chrono::duration_cast< std::chrono::milliseconds >(
             //    std::chrono::system_clock::now().time_since_epoch());
--- a/utils.cpp
+++ b/utils.cpp
@@ -195,7 +195,7 @@ void debugshow(const cv::Mat &m) {
 */
 
 
-void CropGaussianBlur(cv::Mat in, cv::Mat out, int kernelSize, const CircleOutline &outside, const CircleOutline &center)
+void CropGaussianBlur(const cv::Mat &in_, cv::Mat &out, int kernelSize, const CircleOutline &outside, const CircleOutline &center)
 {
     // Warning - this function is extremely sensitive to getting the mask off by even just a few pixels out of all 360k or so.  The mask
     // must be perfect to every last pixel.  If the outer mask zeros some valid pixels we get a turned edge.  Very visible in the 3d view.
@@ -213,6 +213,7 @@ void CropGaussianBlur(cv::Mat in, cv::Mat out, int kernelSize, const CircleOutli
     // we also zero out the input data which probably is already zeroed but we can't take that chance particularly if there were wavefronts averaged
     // because the average feature fills in the center and outside the outer outline.
 
+    cv::Mat in = in_.clone();
     const int height = in.rows;
     const int width = in.cols;
     cv::Mat mask = cv::Mat::ones(height,width,CV_64F);
@@ -257,8 +258,20 @@ void CropGaussianBlur(cv::Mat in, cv::Mat out, int kernelSize, const CircleOutli
 
 
 
-    // these next 3 lines of code are brilliant and from julien asking chatgpt but I get it completely - if someone wants an explanation
-    // I'll have to draw some pictures...  Although I try to explain with words farther below
+    // these next 3 lines of code are brilliant and from chatgpt but I get it completely.  But it's very hard to explain without pictures.
+    // here's an attempt: When you are getting the gaussian blur of a pixel near the edge (near the edge of the mirror or other mask edge) you are adding
+    // up a bunch of points convolved (multiplied) by the kernel shape.  But the pixels outside the mask should be ignored. and because the kernel is
+    // normalized (adds up to 1) you normally don't have to divide by the sum of the kernel values. but you don't want to use the kernel values for points outside
+    // the mirror edge so we set those to zero in the "in" matrix in code above (and you sort of haven't used them because they are all multiplied by zero) but
+    // in this case you need to divide by the kernel values that fall inside the mirror outline.  So we make a mask (with 1s for valid data) and blur it
+    // the same amount and this gives us the proper weighting. In the blurred mask, points far from the edge will all be set to 1 so it won't affect the weighting.
+    // Points just inside the edge of the mirror will have a smaller weighting value. These points near the edge will have a value of the sum of kernel points.  Just
+    // what we need to divide the points by.
+    // This is "mask" in the division below.  To be clear, "mask" will have values less than 1 (greater than 0) near the edge of the mask and is the perfect amount
+    // to divide into "out" to get the properly weighted final result.
+    //
+    // Note the 3 lines of code below can have a very complicated mask with many masked regions and it all still just works and in fact cropGaussianBlur() supports
+    // the inner outline as well as part of the mask.
 
     cv::GaussianBlur(in,out,cv::Size(kernelSize,kernelSize),0);
     cv::GaussianBlur(mask,mask,cv::Size(kernelSize,kernelSize),0);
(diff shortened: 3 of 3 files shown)


simulationsview.cpp
```diff
@@ -35,6 +35,7 @@
 #include <QTextDocument>
 #include <QtMath>
 #include <opencv2/core/core_c.h>
+#include "utils.h"
```
atsju: this triggers me more than it should: why here ? between 2 openCV includes haha.

utils.cpp
```diff
@@ -30,3 +32,287 @@ long showmem(const QString & /*title*/){
 #else
 int showmem(const QString&){return 1000;}
 #endif
+
+
+
+/*
+void debugshow(cv::Mat m) {
```
atsju:
```suggestion
void debugshow(const cv::Mat &m) {
```

utils.cpp
```diff
@@ -30,3 +32,287 @@ long showmem(const QString & /*title*/){
...
+void CropGaussianBlur(cv::Mat in, cv::Mat out, int kernelSize, const CircleOutline &outside, const CircleOutline &center)
+{
+    // Warning - this function is extremely sensitive to getting the mask off by even just a few pixels out of all 360k or so.  The mask
+    // must be perfect to every last pixel.  If the outer mask zeros some valid pixels we get a turned edge.  Very visible in the 3d view.
+    // if it's off the other way you will see some pixels at level 0 outside of the wavefront.  Either error messes up foucault and
+    // ronchi simulations
+
+    // Right now we shrink outer outline by 2 pixels and enlarge inner outline by 1 pixel. As far as I can tell, this is an arbitrary
+    // hack.  These values should at a minimum be in some constant somewhere or better, part of the outline class or maybe we should
+    // get rid of these alltogether?  For now, know that if you mess with it one one place in DFTFringe, you have to mess with it
```
atsju: 😬

utils.cpp
```diff
@@ -30,3 +32,287 @@ long showmem(const QString & /*title*/){
...
+    // in other places as well.
+
+
+    // first we make a mask - with ones where the data is and zeros where there is no data.  So if 0 is black it will be a white donut
+    // using outside and center outlines (often center outline has radius zero so then a disk instead of a donut)
+    // we also zero out the input data which probably is already zeroed but we can't take that chance particularly if there were wavefronts averaged
+    // because the average feature fills in the center and outside the outer outline.
+
+    int height = in.rows;
+    int width = in.cols;
```
atsju:
```suggestion
    const int height = in.rows;
    const int width = in.cols;
```

utils.cpp
```diff
@@ -30,3 +32,287 @@ long showmem(const QString & /*title*/){
...
+                    in.at<double>(y,x) = 0;
+                }
+            }
+        }
+    }
+
+
+
+    // these next 3 lines of code are brilliant and from julien asking chatgpt but I get it completely - if someone wants an explanation
+    // I'll have to draw some pictures...  Although I try to explain with words farther below
```
atsju: put explanation directly here. No need to mention me or chatGPT I suppose

atsju (commented):
sorry for the many comments. Not all require changes I just gone full review mode

surfacemanager.cpp
```diff
@@ -74,7 +74,8 @@
 #include "oglrendered.h"
 #include "ui_oglrendered.h"
 #include "astigpolargraph.h"
-
+#include <chrono>
```
gr5:
```suggestion

```

surfacemanager.cpp
```diff
@@ -430,15 +432,22 @@ void SurfaceManager::generateSurfacefromWavefront(wavefront * wf){
...
             // compute blur radius
             int gaussianRad = 2 * wf->m_outside.m_radius * m_gbValue * .01;
 
             gaussianRad &= 0xfffffffe;
             ++gaussianRad;
 
-            cv::GaussianBlur( wf->nulledData.clone(), wf->workData,
-                              cv::Size( gaussianRad, gaussianRad ),0,0,cv::BORDER_REFLECT);
+            //std::chrono::milliseconds ms = std::chrono::duration_cast< std::chrono::milliseconds >(
+            //    std::chrono::system_clock::now().time_since_epoch());
```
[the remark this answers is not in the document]
gr5: It took me way too long to write that (maybe 15 minutes) and I know I'll need it again. This is the second time I had to time something in the last year and I lost the first method (which was different).  I know I can find it again if I leave it around the gauss blur call.  Not ideal, I know.

utils.cpp
gr5: I want to give credit where it's due.  I never would have come up with blurring the mask but understood it almost immediately when I saw the division on the following line.  I never never would have thought of doing that.  It's just too... outside the box.  It's brilliant really.  I kind of understood almost immediately but had to work through examples in my head to understand that it would work perfectly.

I can move the explanation up though.  I'll do that.

utils.h
```diff
@@ -2,4 +2,7 @@
 #define UTILS_H
 int showmem(const QString &title = "");
 extern double outputLambda; //TODO nothing to do here. But there are many other occurences
+#include <opencv2/opencv.hpp>
+void CropGaussianBlur(const cv::Mat &in, cv::Mat &out, int kernelSize, const CircleOutline &outside, const CircleOutline &center);
```
gr5: @atsju - I can't make it const because I change the "in" matrix.  I zero out the values outside the mirror edge because in some cases (like if you had done averaging) it's possible some of the pixels outside the mirror edge are non zero and that will mess up the gaussian blur.  I assume/hope that if I *don't* make it const and get rid of the "by reference" (the way it was) then it makes a copy, right?  So it's okay if I modify "in", right?

I can read code that uses matrix but it's hard for me to write it because I don't quite get the memory allocation that's going on somewhat invisibly.  But I assume that it makes a temporary copy of the "in" matrix to be used in my cropGaussianBlur() function and then the memory is deleted.  Right?

atsju: no. if you pass by reference without const, it will be in/out and modified outside like a pointer.

I would pass by const reference and explicitly make a tmp copy.
This way reader knows you don't want to modify externally but you need a temporary copy.

utils.cpp
```diff
@@ -30,3 +32,287 @@ long showmem(const QString & /*title*/){
 #else
 int showmem(const QString&){return 1000;}
 #endif
+
+
+
+/*
```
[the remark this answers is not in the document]
gr5: I'm leaving it commented out.  If I remove the function I think I should remove it from the .h file as well.

utils.h
atsju: OK I spoke with friend @hish15 It's not that much about https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rf-in but rather "P.5 Express intent". 
This function does not modify the object outside so it needs to pass const ref. If you need a local copy: then make a local copy. Doing in a different way is unclear because the parameter passing is also doing the local copy you need.

Note; you should also apply https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rf-out for out parameter. Return `out` instead passing it.

gr5: I did the "out" that way as that is how the old cv::gaussianBlurr() does it.  I was just copying the way they did it.

I can't get the const to work.  I get compile errors.  I don't understand the errors.  Still working on this.

Keep in mind that before, it *did* modify "in".  I even had to send in a clone in one place in the code.  Kind of like:
CropGaussianBlur(wf->nulled_data.clone(),out,...

Now that line of code I got rid of the "clone()" as it was duplicative and put the clone inside the function.  But the compiler complains of undefineded reference to my function.  sigh.  I modified it in both utils.h and utils.cpp.

gr5: found the problem.  I saw you added the & in the .h file to "in" but didn't notice "out" and I only fixed the "in" in the .cpp file.

Will commit the fix soon.

gr5:
yes about the copy.  I used ".clone()" function typically which creates a fresh matrix and copies all the date.  Or you can use copyTo() if both matrixes exist.  But using operator= I don't know what it does.  Probably just does a pointer?

githubdoe:
Yes "=" and passing by value give a reference to the data.  It is discussed in OpenCV matrix doc under memory management I think.

That is why you will find several places where I do a clone() operation.  

It may confuse the warning message maker however.  I don't know but I don't think it needs special treatment when passed by value or reference.

surfacemanager.cpp
```diff
@@ -359,8 +359,9 @@ void SurfaceManager::generateSurfacefromWavefront(wavefront * wf){
                 gaussianRad &= 0xfffffffe;
 
                 ++gaussianRad;
-                        qDebug() << "Blurr" << gaussianRad;
-                    cv::GaussianBlur( wf->nulledData.clone(), wf->workData,
+
+                qDebug() << "Blurr" << gaussianRad;
+                cv::GaussianBlur( wf->nulledData.clone(), wf->workData,
```
atsju: @gr5 This is the last cv::GaussianBlur. why this one doesn't benefit from the new function

gr5:
I think adding the const pass by reference helps show the intent of the function as well.  A kind of self documentation that the function won't mess with the "in" matrix but will be modifying the "out" matrix (which is also now passed by reference but without the const).

Although calling them "in" and "out" also did that somewhat and even then I violated that by modifying "in" before these code review changes.

surfacemanager.cpp
atsju: If it can be improved then add an issue I suppose. I'm approving this as the code is ok.

PR #274, opened 2025-11-14
add some lvl2 clazy checks and ignore external libs

### LINKED ISSUE #273: Enable some lvl2 clazy warnings
For example I fixed all the pass by reference VS value. The warning need to be enabled for new code to trigger it.

atsju:
fix #273


gr5 (approved):
This looks fine.  I tested the star test movie feature as well.  Still works as it did before.

PR #278, opened 2025-11-19
disable windows darkmode support

### LINKED ISSUE #264: Add mode setting to DFTFringe preferences.
DFTFringe looks terrible if a windows 11 machine has dark mode set.

AI says if you do the following you can change the mode.  
So perhaps add a setting in DFTFringe that enables the mode. 

Here is the code to force it at startup.  

    #include <QApplication>
    #include <QProcessEnvironment>

    int main(int argc, char *argv[]) {
        QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
        env.insert("QT_QPA_PLATFORM", "windows:darkmode=0"); // Force light mode
        QProcessEnvironment::setSystemEnvironment(env);

        QApplication app(argc, argv);
        // ... rest of your application code
        return app.exec();
    }

atsju:
fix #264

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -92,7 +92,7 @@ static void my_terminate_handler() {
 static void myQtMessageOutput(QtMsgType type, const QMessageLogContext &/*context*/, const QString &msg)
 {
     const std::string localMsg = msg.toStdString();
-    
+
     switch (type) {
     case QtDebugMsg:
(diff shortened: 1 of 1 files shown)


atsju:
I did miss the fact that there were build warning annotations in review pane. I will need to fix this !

PR #279, opened 2025-11-19
[DRAFT FOR TESTS] adding clang tidy comments on modified code

atsju:
This is a new github action for code quality. It should annotate code based on advanced checks (that we can fine tune) https://clang.llvm.org/extra/clang-tidy/checks/list.html

I made it so that only new code will be annotated and we live with older code as it is.

I'm working on this and not on some other more important features because I want to try it out and eventually reuse it at work.

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -12,6 +12,9 @@ env:
 
 jobs:
   build-linux:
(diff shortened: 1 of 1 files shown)


main.cpp
atsju: test

PR #280, opened 2025-11-20
fix compilation warnings introduced in #278

atsju:


gr5 (approved):
Looks good.  I downloaded and tested in dark mode and it works as intended (shows light mode app over a dark mode desktop).

PR #282, opened 2025-11-23
New release v8.2.0

gr5:
would be nice to have circular contour plot if that's almost ready in which case we should add that to this list

### DIFF
--- a/RevisionHistory.html
+++ b/RevisionHistory.html
@@ -18,9 +18,11 @@
             <li>Darkmode disabled (was a problem for people with win11/win12 who set windows to dark mode)</li>
             <li>Annulus feature help text improved</li>
 			<li>New shortcut keys:</li>
-            <li>e - 'edge' does instant edge zoom when positioning outline (f 'full' goes back to full view; click 'help' for more shortcuts)</li>
-            <li>i - 'import' loads latest igram from folder set in preferences. Very handy feature!</li>
-            <li>s - 'save' saves current wavefront</li>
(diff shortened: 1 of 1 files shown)


atsju:
> would be nice to have circular contour plot if that's almost ready in which case we should add that to this list

I don't know when I will have time to fix properly. 
Worst case we do 8.2.1

RevisionHistory.html
```diff
@@ -10,6 +10,20 @@
     <h1>DFTFringe Version History</h1>
     <ul>
 
+        <li>Version 8.2.0</li>
+        <ul>
+            <li>Gaussian blur no longer turns edges (and other artifacts) -- zernike blur no longer needed for foucault/ronchi/star test simulations</li>
+            <li>Profile plot can maximize once again along with ability to change units among inches, mm, percent.</li>
+            <li>Zonal Percent correction graph from 7.3.1 moved to the context menu (right click) of the profile plot</li>
+            <li>darkmode disabled (was a problem for people with win11/win12 who set windows to dark mode)</li>
+            <li>annulus feature help text improved</li>
```
atsju:
```suggestion
            <li>Darkmode disabled (was a problem for people with win11/win12 who set windows to dark mode)</li>
            <li>Annulus feature help text improved</li>
```

Win 12 ?? did you mean 10/11 ?

atsju (commented):
You also fixed a minor bug with 'h' in #277

PR #284, opened 2025-11-25
Fix execution on Linux. Brocken by windows only darkmode args in #278

Lucassifoni:
Hello,
I do not know anything about QT nor windows development in general, so this might be wrong. To build and run on linux without passing a platform flag, I have been forced to add those preprocessor directives. This might not be the correct way to fix this, and it might even be a documentation / README issue ?

Let me know if more information is needed to determine if this is specific to my system or not. Hope I am not wasting anyone's time here.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -123,36 +123,32 @@ static int myCvErrorCallback( int /*status*/, const char* /*func_name*/,
 
 int main(int argc, char *argv[])
 {
-    // DFTFringe doesn't have a good darkmode palette
-    // one could call "DFTFringe.exe -platform windows:darkmode=1" to disable dark mode (except for app borders)
-    // Following code adds the platform argument programmatically
-
-    // Create a new argv array with existing args plus platform args
-    #ifdef Q_OS_WIN
-     int newArgc = argc + 2;
+    #ifdef _WIN32
+        // DFTFringe doesn't have a good darkmode palette
+        // one could call "DFTFringe.exe -platform windows:darkmode=1" to disable dark mode (except for app borders)
+        // Following code adds the platform argument programmatically
+        // Create a new argv array with existing args plus platform args
+        int newArgc = argc + 2;
+        std::vector<std::string> args;
+        args.reserve(newArgc);
+        // Copy existing arguments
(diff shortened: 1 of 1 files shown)


atsju:
You are not wasting time. Feedback is welcome. 

Build on linux was passing on continuous integration so I'm suprised this change is needed to build.
But I haven't tried execution.  I will execute and let you know.

Lucassifoni:
DFTFringe built fine without this change. But it did not run.
```
lucas@lucas-ThinkPad-X270-W10DG:~/DFTFringe-8.2.0$ ./DFTFringe
qt.qpa.plugin: Could not find the Qt platform plugin "windows" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
Available platform plugins are: wayland, wayland-egl, vkkhrdisplay, minimal, offscreen, eglfs, minimalegl, linuxfb, vnc, xcb.
Abandon (core dumped)
lucas@lucas-ThinkPad-X270-W10DG:~/DFTFringe-8.2.0$ QT_QPA_PLATFORM=xcb ./DFTFringe 
qt.qpa.plugin: Could not find the Qt platform plugin "windows" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
Available platform plugins are: wayland, wayland-egl, vkkhrdisplay, minimal, offscreen, eglfs, minimalegl, linuxfb, vnc, xcb.
Abandon (core dumped)
```

Lucassifoni:
> > Also, for some reason, the README instructions did not work as-is on my distro. The second qmake invocation did not find a project autonomously, I had to point it to DFTFringe.pro Not sure is this is system-specific or if the instructions should be updated, since I've seen it works fine in the CI.
> 
> Indeed README is correct and works on CI and my computer. I don't know how to address this.

I can maybe add a conditional to the README in another PR ?

Ex : 

*Note : if the second invocation of qmake, at the root of the source directory does not succeed, you might need to explicitly pass the project file, `/usr/lib/qt6/bin/qmake DFTFringe.pro`.*

PR #288, opened 2025-11-27
fix crash in View/Show statistics

### LINKED ISSUE #283: 8.2.0 crashes in View/Show statistics
https://groups.io/g/Interferometry/message/36742

atsju:
fix #283

### DIFF
--- a/wftstats.cpp
+++ b/wftstats.cpp
@@ -257,31 +257,22 @@ void wftStats::computeWftRunningAvg( QVector<wavefront*> wavefronts, int ndx){
     outliersInner.clear();
     outliersOuter.clear();
     // normalize the size to the most common size
-    int last = wavefronts.length();
-    QHash<QString,int> sizes;
-    for (int i = 0; i < last; ++i){
-        QString size = QString("%1 %2").arg(wavefronts[i]->workData.rows).arg(wavefronts[i]->workData.cols);
-        if (sizes.contains(size)) {
-            ++sizes[size];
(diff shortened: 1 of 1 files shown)


gr5:
This fix is fine and I tested it but these lines of code a little further down...  YIKES.

https://github.com/githubdoe/DFTFringe/blob/098dd331550ceba7ad4ac721c0e4db651f0188a9/wftstats.cpp#L286-L302

It appears that it is summing up all the wavefronts into a matrix called sum which is great but **inside the loop** it is calculating the average.  So if you have 100 wavefronts (Yordan often has 100 wavefronts) and looking at the statistics it calculates the average 100 times (not 99, 100).  And calculates the standard deviation 100 times.  That code (lines 294-297) should be outside the for loop. I'm pretty sure.

Also one could just put only the zernike's in a table instead of creating an entire wavefront matrix and do the std deviation on just the table of zernikes (once for each of the 49 or so zernikes).

Anyway, I don't want to mess with it as it would need careful testing I think.

githubdoe:
Just a reminder of the history of that feature. We knew that if you can make the noise random (air currents sort of are)  then if you average enough wave fronts samples to randomize the noise you can remove the noise.  The trick is knowing when you have enough samples.  That depends on the number of samples and the noise.  So I wanted a way to show when enough samples had been taken.  That feature average each sample as a running average and when/if the STD or other values you watch settle into the lowest value you have enough samples.

PR #291, opened 2025-11-30
fix openCV version in readme

atsju:


atsju:
draft as I need to check something regarding libopencv_flann4120.dll

atsju:
`$$PWD` expands to current .pro file location. It technically changes nothing but without it, build wasn't working today on the new computer. 
This change should not affect you. It should not be required but at the same time it shouldn't harm anyone.

PR #295, opened 2025-12-03
Right side tabbed tools can be much smaller (250pix versus 524pix)

### LINKED ISSUE #289: UI right bar to adjust for Dale and other screen
It should look similar to that :
Image

Shall be usable on a 17" screen with 100% scaling (most people) and 125% scaling (Dale uses 250% with 2x2 times resolution so this is equivalent)

See #286 for discussion details

gr5:
This will greatly help the app fit on smaller displays or displays where scaling is > 200%.

This was a new issue for Dale when we switched to QT6 but also has been an issue with people whose screen width is < 1280 pixels.

This will help everyone, even people with large displays as you can choose to allocate less space for the right side tool area and give more display area to other things.

I was able to achieve this by simply changing the minimum widths to 250 for all the .ui files that have to fit in that right pane.  The gui editor in QT was very helpful by showing what it would look like as I made the tools wider and thinner.

temp_gif2

Fixes #289

### DIFF
--- a/contourtools.ui
+++ b/contourtools.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>451</width>
+    <width>250</width>
     <height>497</height>
    </rect>
   </property>
@@ -18,7 +18,7 @@
   </property>
   <property name="minimumSize">
    <size>
-    <width>451</width>
+    <width>250</width>
     <height>411</height>
    </size>
   </property>
--- a/mainwindow.ui
+++ b/mainwindow.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>1065</width>
+    <width>667</width>
     <height>905</height>
    </rect>
   </property>
@@ -18,7 +18,7 @@
   </property>
   <property name="minimumSize">
    <size>
-    <width>640</width>
+    <width>250</width>
     <height>640</height>
    </size>
   </property>
@@ -58,8 +58,8 @@
     <rect>
      <x>0</x>
      <y>0</y>
-     <width>1065</width>
-     <height>19</height>
+     <width>667</width>
+     <height>22</height>
     </rect>
    </property>
    <widget class="QMenu" name="menuFiles">
@@ -161,7 +161,7 @@
    </property>
    <property name="minimumSize">
     <size>
-     <width>475</width>
+     <width>250</width>
      <height>791</height>
     </size>
    </property>
@@ -430,7 +430,7 @@ QPushButton:hover{rgb(85, 255, 255)}</string>
             </property>
             <property name="minimumSize">
              <size>
-              <width>74</width>
+              <width>82</width>
               <height>0</height>
              </size>
             </property>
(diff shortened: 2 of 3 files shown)


githubdoe:
Wonderful.   Good work.  I think you can make it even smaller and let the user decide when it gets too small.  I just tried 100 and it was still ok.

gr5:
100pix is too narrow I think.  Looks like this in qt creator.  Again, this is "minimum" width setting - not sure if that's what you played with.

image

Also realize you have to worry about all 4 tabs.  And already the first tab is a bit tight.  This is how it looks for me at 250px (qt6).  Note how the "edge zoom" checkbox is cut off a bit and a few other things.

image

atsju:
I tested d8acf43. Looks good. However at 125% GUI was not responding very well and I wasn't able to resize as smooth as I wanted. Only 2 possible sizes like it's not analog slidebar. I don't know if you experience same : 


https://github.com/user-attachments/assets/39c89f9a-15b3-4cae-8eae-645bc21ac8eb


I'm not against changing minimum size to a very small value as long as default size displays OK.

gr5:
That weird behavior - it does that in the older version also.  You don't have to go to 125%.  You just have to make the application width a little too small.  I don't know how small exactly.  I tried the version on github from this pr and set to 125% at full screen I was fine but when I shrunk the width of the app a bit it does the same thing yours does.

I don't know how to prevent that weirdness.  But basically it happens when there isn't enough room horizontally in the main window of the application.  I don't know what sets this.


I think a better test is to not touch that right-side-size-adjuster area and instead make the entire app get smaller and larger.  If you do that you can better see where the minimum size comes to play.

I think it's fine as is.

gr5:
Dale, where are we with this?  I made it about half the minimum width and I could go much smaller but it gets less usable pretty quickly.  Can I stick with 250 as the minimum?

gr5:
7 posts above this one there is a link.  Just download that, install it, and try it out.

Or do you want to build it or edit it yourself?  The code changes are tiny - basically 4 lines of code but I did it with the ui editor thing.  I can show you a quick video of what to change in the ui editor.  Would that help?  maybe I'll do that.

Also you can download this particular branch if you have the github cli with this command line command:
gh pr checkout 295

But you have to install the github cli and you have to make sure all your edits are saved or commited (if you have any edits since your last big pull request a few weeks ago).

I'll do a video.  I might have to do it in a few hours as I think dinner is in 10 minutes.

githubdoe:
Thank you for the video.  It helped but did not fix my problem.  I still can't get the width to too small.  I have isolated it to the section called interferogram tools (on mainWindow)  by dragging the header of the control so that it is no longer docked.  All others also dragged will reduce in size but that one will not even when I have minimum set to 100 and dragged it left on the UI editor.  Oh, well.  Carry on.

gr5:
Dale do you want to approve this so I can merge it?  I think Julien is waiting for you to approve it as the code change is of course pretty basic.  It's more about is this an improvement to the program or do we need to go farther.

I don't want to go too far as not everyone will be able to find how to use the mouse to make the tools area larger so I want it to shrink to a value that is still usable (just barely usable).

PR #296, opened 2025-12-06
adding clang-tidy review for pull requests

atsju:
This is a new github action for code quality. It should annotate code based on advanced checks (that we can fine tune) https://clang.llvm.org/extra/clang-tidy/checks/list.html

I made it so that only new code will be annotated and we live with older code as it is.

This is an example of what it would look like if it was active on #263 
https://github.com/atsju/DFTFringe/pull/51/files#diff-06bf2ef5cae34afe079d755885471c5d1374ff88dd3384491e1f290d2c72eb73
@gr5 @githubdoe  let me know what you think.

Not all remarks are to be fixed but some of them are just excellent (https://github.com/atsju/DFTFringe/pull/51/files#r2595159192)
Some are really good practice (override, nullptr, naming 3 chars mini, magic numbers, ...)

However some fixes are only C++11 or C++17 compliant. So it would be easier if we do not have Qt5 to maintain (especially Dale version which I'm not even sure is C++11). Anyway, we can  still enable it, just remember to look for incompatibility before blindly accepting any proposed changes.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,6 @@
+Checks: >
+  clang-analyzer-*
+  bugprone-*
+  performance-unnecessary-*
+  modernize-use-nullptr
+  modernize-use-nodiscard
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -46,6 +46,7 @@ jobs:
           style: '' # we don't want check from clang-format
           lines-changed-only: 'true' # Only lines in the diff that contain additions are analyzed. Avoid noise from existing code.
           thread-comments: 'update'
+          tidy-checks: '' # rely solely on .clang-tidy file
           tidy-review: true
           passive-reviews: true
           ignore: 'bezier|boost|SingleApplication|spdlog|zernike|moc_*|ui_*|qwt*'


atsju:
> I disagree pretty much with all of the clang-tidy suggestions. Here's an example. Complain about "h" variable name too short but it's kind of self documenting here: `int h = c->height();`

I partially disagree with the example you choose. 
It would totally made sense to name the variable "height" instead h where you will need to read affectation to understand. My coding style in this PR is mainly chatGPT unfortunately.

> I prefer Dale's and Julien's coding styles over the clang-tidy suggestions.
> 
> So this would run when I do a PR and then I could go in and fix some of them if I wanted to or ignore them? I think this will make code reviews more difficult if I have to skip over all the clang-tidy notes.

Exactly. And I agree 40 comments will make it slightly more difficult... But some are good to take.

> I think these tools make more sense before I do a commit. Tools that I should be running on my dev environment.

You can https://doc.qt.io/qtcreator/creator-clang-tools.html it's clang tidy. 



In the end I will follow your choice. This is a proposal and I know it's very disruptive and noisy. I'm OK if we don't take this. 
I personally prefer to enforce during PR.

atsju:
Ready for review. Look at #300 for result.

atsju:
I'm merging this. 
If it becomes annoying at some point, we can tune parameters.

Also, the .clang_tidy file is automagically recognized by QTcreator so it should help. You might see some warnings from untouched code locally but there are not that much left with these settings.

PR #300, opened 2025-12-14
[DO NOT MERGE] PR shoving output of clang-tidy in typical case

atsju:
what would it show in #263

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,6 @@
+Checks: >
+  clang-analyzer-*
+  bugprone-*
+  performance-unnecessary-*
+  modernize-use-nullptr
+  modernize-use-nodiscard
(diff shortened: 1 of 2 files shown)


atsju:
OK here we are. Only one warning by clang-tidy on my PR.
Even better: it's 100% relevant. I was concerned by line 2749 ` ContourPlot *plot =new ContourPlot(0,0);//m_contourPlot; //TODO leaking ? ` and wrote the comment before looking at clang-tidy outputs.

I recommend merging #296 (not this one) as it should not be too noisy and still makes good review.

as I explained there is a very tiny bit of warning leaking on unmodified code: https://github.com/githubdoe/DFTFringe/pull/296#issuecomment-3622109039
>>point 3: About half of the clang comments were on code that you hadn't modified. It seems maybe it was either all code in the modified file or it was code near the edits. I'm not sure which.
>
>What it does basically is a diff. If it shows in diff then it's reviewed. If you look again you will see most comments are on modified code (even if modification is only a space change on this line).
>Very few are indeed on unmodified code but very near. This is from how diff works.
>I confirm the whole file is not reviewed. This has been tested and verified.

PR #307, opened 2025-12-23
fix some spellings in settings

atsju:


atsju:
> These are great. I see so many spelling errors when going over DFTF. I should fix them more often.
> 
> I read a lot of novels. About 1 per week (40 per year - I keep track). Because of that spelling errors stand out visually to me.

I did not fix the red underlined in pictures. Kindly confirm those have wrong capitalisation and I will fix them.

gr5:
The titles of books and movies tend to have almost every word capitalized other than {the,of,a}.  But the examples you underline above are all backwards capitalization as you probably suspect.

atsju:
Our messages crossed. It's clear thank you.

I want to do some UI fixes but it might come after the release

githubdoe:
I tend to capitalize words I want to emphasize in titles .  I view text that document user parameters as titles. 

 So for example "Mask" is the thing being displayed.   It is capitalized to separate it from the other text.  These are titles not pros.  
Now having said that I'm not always consistent.

gr5:
> I tend to capitalize words I want to emphasize in titles .

Bold or Italic would be better for that although I don't know if that's easy to do in QT.

gr5:
Dale, to be clear, Julien moved this button as shown.  Previously it was just above "apply" button.  I think it is fine in either location and have no preference.  Just wanted you to be aware.  This is in addition to some spelling errors.  So far, this PR looks good to me.

image

atsju:
@gr5 @githubdoe I have done a small video with further review of "Settings". https://youtu.be/Lb3vSjkuqo8
Let me know what you think. I don't want to invest too much work if you don't like the outcome.

Recap:
- remove "apply"
- pack lists that are too spaced
- render everything fixed size to avoid all issues linked to resizing

atsju:
Yes I have played a little with Qt designer (the UI tool in qt creator). I know it can be frustrating. I don't plan to edit UI files programmatically.

Do you prefer the simple solution of fixed size or should I go the long path making things resizable ? I mean it's nice but it requires more tests and work to bring little value.

I will remove the lens things if nobody uses it. Shall we ask on groups.io before removing ?

gr5:
Did you still mean for this to be in draft mode?  Or do you have more to do?

PR #309, opened 2025-12-23
Average profile

githubdoe:
Added different way to create average profile and display it with right click menu.
Added other things to right click menu of profile plot.
Added right click menu of Ronchi image to create image of multiple ronchi images.
Added if you move the avg or single profile it will also move the 16 diameters profiles once they are displayed.
Profile offset will happen whenever you select different wave fronts.

### DIFF
--- a/foucaultview.cpp
+++ b/foucaultview.cpp
@@ -490,7 +490,7 @@ void foucaultView::generateBatchRonchiImage(const QList<wavefront*>& wavefrontLi
     layout->addLayout(btns);
 
     // Connect the comparison trigger
-    connect(compareBtn, &QPushButton::clicked, [=, &previewDlg]() {
+    connect(compareBtn, &QPushButton::clicked, &previewDlg, [=, &previewDlg]() {
         // [=] copies individualRonchis and names so they stay
         // valid even after generateBatchRonchiImage() returns.
         if (individualRonchis.size() >= 2) {


gr5:
Oh.  There are merge errors.  This is related to my recent PR where I added the "origin" enum to a few functions.  I can do the merge if you want Dale.  I also marked this PR as draft.  Search this web page for "draft" to see the link to switch it back.

githubdoe:
Those were my temporary changes to keep the dialog from popping up until you had your permanent fix.

githubdoe:
The stop auto invert is leftover from much earlier time when there was no invert dialog.  It may no longer be needed.

githubdoe:
Profile offset reset will happen whenever you select different wave fronts.

githubdoe:
I have pulled the code and it compiles.  I see the square contour plot but do not see the vertical blue bar for the right side.  It will still drag that control however.

I just realized an addition I want to add.  That is overlay one ronchi with another to see how they are different.  I bet AI can help me with that as well.

gr5:
Also tomorrow I'll be quite busy with visiting family so may not notice anything in my emails nor anything github related. 😄

atsju:
I think I fixed the build. I'm more busy than expected so I don't know how much time I have available.

Dale, you will need to pull the branch to avoid conflicts. As George explained: kindly avoid pushing new features to this branch until it's merged. It should probably already have been split into 2 or more PR but I don't want to refrain you developping so we will just take care of it as it is.
I will make a more formal code review as soon as I have more time. 

Happy Xmas to you 2 :)

atsju:
> I would remove the ifdef notnowold. But then you would get my new feature as well.

I'm not sure to understand. 
If you are talking about the warning, the ifdef is not the cause. The  problem is you have 2 macro after ifdef instead one. Compiler might not complain but it's not language compliant.

If you want to remove it and talk about "no new feature", go ahead and remove, your new feature is aleady in the code and compiled by default. Removing the old code is part of this PR. What we want to avoid is getting more unrelated features.

Just remember to pull the branch now to avoid conflicts. As George and I pushed changes

foucaultview.cpp
```diff
@@ -74,10 +74,27 @@ void foucaultView::showContextMenu(QPoint pos)
...
 }
+
+void foucaultView::showSelectedRonchiImages(){
+
+    surfaceAnalysisTools *saTools = surfaceAnalysisTools::get_Instance();
+    QList<int> list = saTools->SelectedWaveFronts();
+
+    QList<wavefront*> wfs;
+
+    for (int i = 0; i < list.size(); ++i){
```
atsju: I like to use size_t for arrays and lists as this is what size() returns. 
low importance
Shall compile same.
```suggestion
    for (size_t i = 0; i < list.size(); ++i){
```

foucaultview.cpp
```diff
@@ -152,6 +169,328 @@ QVector<QPoint> scaleProfile(QPolygonF points, int width,
...
+        int ppl = std::max(1, (int)((0.5 / lpi_val) / pixwidth));
+        int start = (size / 2) - (ppl / 2);
+        bool even = ((start / ppl) % 2 == 0) ^ s.clearCenter;
+        int roffset_start = ppl - (start % ppl);
+
+        for (int y = 0; y < size; ++y) {
+            int line_no = 0, roffset = roffset_start;
+            for (int x = 0; x < size; ++x) {
+                if (((even && (line_no % 2 == 0)) || (!even && (line_no % 2 != 0)))) mask.at<double>(y, x) = 1.0;
+                if (++roffset >= ppl) { line_no++; roffset = 0; }
```
atsju:
```suggestion
                roffset++;
                if (roffset >= ppl)
                {
                    line_no++;
                    roffset = 0;
                }
```

foucaultview.cpp
```diff
@@ -152,6 +169,328 @@ QVector<QPoint> scaleProfile(QPolygonF points, int width,
...
+    settings.lateralOffset  = this->lateralOffset;
+    // Note: outputLambda should be defined/accessible in your scope
+    settings.outputLambda   = 550.0;
+
+    // 2. Call the refactored static engine for both images
+    QImage ronchiImg = generateOpticalTestImage(OpticalTestType::Ronchi, m_wf, settings);
+    QImage foucaultImg = generateOpticalTestImage(OpticalTestType::Foucault, m_wf, settings);
+
+    // Store for potential saving/external access
+    m_foucultQimage = foucaultImg;
```
atsju:
```suggestion
    m_foucaultQimage = foucaultImg;
```

foucaultview.cpp
```diff
@@ -152,6 +169,328 @@ QVector<QPoint> scaleProfile(QPolygonF points, int width,
...
+
+    // 3. UI Painting Helper (to avoid duplicating the Painter logic)
+    auto paintAndDisplay = [&](QLabel* label, QImage img, double offsetValue) {
```
atsju:
```suggestion
    auto paintAndDisplay = [&](QLabel* label, const QImage& img, double offsetValue) {
```
if you need a copy of `img`, do it explicitely in the code for readability. If you don't then this will just improve performance by removing an unnecessary copy

foucaultview.cpp
```diff
@@ -152,6 +169,328 @@ QVector<QPoint> scaleProfile(QPolygonF points, int width,
...
+    // Draw separator line
+    painter.setPen(QPen(Qt::gray, 2));
+    painter.drawLine(10, headerHeight - 15, canvas.width() - 10, headerHeight - 15);
+
+    // 6. Loop through and Render Ronchi Patterns
+    painter.setFont(QFont("Arial", 11, QFont::Bold));
+    QFontMetrics fm(painter.font());
+
+    for (int i = 0; i < count; ++i) {
+        wavefront* currentWf = wavefrontList[i];
```
atsju:
```suggestion
for (wavefront* currentWf : wavefrontList) {
```
"range for" are realy good for such cases. No count, better readability

plotcolor.cpp
```diff
@@ -1,7 +1,7 @@
 #include "plotcolor.h"
 const char *plotColors[]=
-{   "Red",
-    "Black",
+{   "BLack",
```
atsju:
```suggestion
{   "Black",
```

profileplot.cpp
```diff
@@ -611,137 +695,137 @@ void ProfilePlot::populate()
...
-            list.clear();
-            for (int i = 0; i < wfs->size(); ++i){
-                list << i;
-            }
-        }
+        // if no wave front was selected then use the last one
+
         for (int i = 0; i < list.size(); ++i){
-            QStringList path = wfs->at(list[i])->name.split("/");
+            wavefront* wf = wfs->at(list[i]);
```
atsju: "range for"

foucaultview.cpp
```diff
@@ -152,6 +169,328 @@ QVector<QPoint> scaleProfile(QPolygonF points, int width,
...
+    cv::Mat mask = cv::Mat::zeros(size, size, CV_64FC1);
+    cv::Mat sourceSlit = cv::Mat::zeros(size, size, CV_64FC1);
+    int hx = (size - 1) / 2 + s.lateralOffset;
+    double pixwidth = s.outputLambda * 1.E-6 * (0.5 * md->roc / md->diameter) * 2. / (25.4 * actualPad);
+
+    if (type == OpticalTestType::Ronchi) {
+        double lpi_val = s.lpi * (s.useMM ? 25.4 : 1.0);
+        int ppl = std::max(1, (int)((0.5 / lpi_val) / pixwidth));
+        int start = (size / 2) - (ppl / 2);
+        bool even = ((start / ppl) % 2 == 0) ^ s.clearCenter;
```
atsju:
```suggestion
        bool even = ((start / ppl) % 2 == 0) != s.clearCenter;
```

^ is a bitwise operator.

atsju (commented):
I did not test. Only code review.

autoinvertdlg.ui
```diff
@@ -137,6 +137,48 @@
...
+          <bool>true</bool>
+         </property>
+        </widget>
+       </item>
+       <item>
+        <widget class="QRadioButton" name="DoNotAskorInvertRb">
+         <property name="text">
+          <string>Don't ask or invert</string>
+         </property>
+        </widget>
```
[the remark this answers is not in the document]
githubdoe: Yes that code is my temp fix.  It should not be merged.

foucaultview.cpp
githubdoe: Yes I think in this case ^ is correct.  This was code generated by AI  a clever trick to test the int result being 0 or 1 along with the test for a clear center setting.

foucaultview.cpp
githubdoe: ok

foucaultview.cpp
```diff
@@ -152,6 +169,328 @@ QVector<QPoint> scaleProfile(QPolygonF points, int width,
...
+    s.lpi            = ui->lpiSb->value();
+    s.gamma          = ui->gammaSb->value();
+    s.slitWidth      = ui->slitWidthSb->value();
+    s.useMM          = ui->useMM->isChecked();
+    s.movingSource   = ui->movingSourceRb->isChecked();
+    s.knifeOnLeft    = ui->knifeOnLeftCb->isChecked();
+    s.clearCenter    = ui->clearCenterCb->isChecked();
+    s.heightMultiply = this->heightMultiply;
+    s.lateralOffset  = this->lateralOffset;
+    s.outputLambda   = 550.0;
```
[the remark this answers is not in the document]
githubdoe: All values set by the user using the UI but can now be used by routines without access to the UI.  The AI uncoupled the image generation from the class that it is in.  Users no longer have to have access to the class instantiation (the singleton) to get a ronchi image.  That may be useful for things like report generation.

foucaultview.cpp
```diff
@@ -152,6 +169,328 @@ QVector<QPoint> scaleProfile(QPolygonF points, int width,
...
+    settings.useMM        = ui->useMM->isChecked();
+    settings.movingSource = ui->movingSourceRb->isChecked();
+    settings.knifeOnLeft  = ui->knifeOnLeftCb->isChecked();
+    settings.clearCenter  = ui->clearCenterCb->isChecked();
+
+    // Pass the class members that aren't driven by UI widgets
+    settings.heightMultiply = this->heightMultiply;
+    settings.lateralOffset  = this->lateralOffset;
+    // Note: outputLambda should be defined/accessible in your scope
+    settings.outputLambda   = 550.0;
```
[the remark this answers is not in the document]
githubdoe: That value and all values in the new settings structure generated by AI are values that will not be used by current code.  The real values come from foucaultview UI and the 550 is the output wavelength specified in the preferences under the configuration menu.  550 is the wavelength of green light in nanometers.   I'm impressed that AI figured out a value for [t there even though it did not have access to the code where that value is set.  

AI generated those default values which would be used only if we ever write code outside the foucaultview.cpp file to generate ronchi images where we don't have access to the foucaultview UI.

profileplot.cpp
```diff
@@ -547,9 +537,103 @@ void ProfilePlot::make_correction_graph(){
...
+ * returning a vector where index r is the average height at radius r.
+ * It assumes the input matrix is of type CV_64FC1 (double).
+ * * @param matrix The 2D matrix (cv::Mat, expected CV_64FC1).
+ * @param center The center point of the circular surface (cv::Point).
+ * @param radius The radius of the circular surface in pixels.
+ * @return A std::vector<double> where index r contains the average height at integer radius r.
+ */
+std::vector<double> compute_average_radial_profile(
+    const cv::Mat& matrix,
+    const cv::Point& center,
```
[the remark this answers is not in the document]
githubdoe: why?  This code will not modify center.

githubdoe:
I can not make any changes to that code because my current code has added functionality in it that both you and George said not to push.

gr5:
The multiple-ronchi feature is nice but whether I build it QT5 or QT6 I get the same issue with these ok/cancel buttons:

image


image

gr5:
> I can not make any changes to that code because my current code has added functionality in it that both you and George said not to push.

Well there are 2 options.

option 1, grab the changes we made so far and merge them in.  I don't think we changed much so far.  I merged in master - okay that's a big change but there were no merge conflicts reported by git and there was one small edit (one line of code changed) to get it to compile after the merge.  Also Julien made a tiny change to get it to compile in QT6 (he added a single include).

option 2, instead of pushing your current working copy to git (averageProfile branch) you can switch to a brand new branch and push that out so that it doesn't affect this PR.  That way your working branch would be exactly as it is now.  But eventually we want you to merge in master and the 2 changes I just mentioned in option 1.

I'm not sure what Julien and Dale prefer.

githubdoe:
I wondered about those buttons and thought that it was a Qt5 thing.  So did not play with them.  There are just too tiny for me to read the text.  Looks like the text does not even make through the black.   So perhaps there are two problems.  
1. Tiny text and buttons that I get with Qt5.  
2. Background color and text color.   Maybe just remove the black background.

githubdoe:
I have asked AI about the problem and it says there must be a style sheet somewhere that is setting the global button style.  To experiment add this code just before the compare dialog is opened.  

It works and now I can read the buttons.  So I need to look for the overriding style sheet.

Here is the code that makes the buttons readable:

QString buttonStyle = "QPushButton { background-color: palette(button); color: palette(button-text); border: 1px solid gray; padding: 4px; }";
compareBtn->setStyleSheet(buttonStyle);
saveBtn->setStyleSheet(buttonStyle);
cancelBtn->setStyleSheet(buttonStyle);

githubdoe:
AI gives two fixes.  

image

gr5:
Oh it was indeed easy.  Took 7 minutes (didn't feel that long!).  Dale added ronchicomparedialog to only one of 3 project files.  Hopefully it will compile with no errors and then I'll look it over.

githubdoe:
Sorry I keep forgetting to add new files to other .pro files.  Thanks

gr5:
I played with the ability to diff two ronchi's.  It shows the changes in red and green and is very cool - it trains your eyes where to look to see subtle changes to the mirror in the ronchigram.  Nice feature.

plotcolor.cpp
gr5: commiting this

foucaultview.cpp
```diff
@@ -152,280 +169,351 @@ QVector<QPoint> scaleProfile(QPolygonF points, int width,
...
+    QPushButton *cancelBtn = new QPushButton(tr("Close"));
+
+    btns->addWidget(compareBtn);
+    btns->addStretch();
+    btns->addWidget(saveBtn);
+    btns->addWidget(cancelBtn);
+    layout->addLayout(btns);
+
+    // Connect the comparison trigger
+    connect(compareBtn, &QPushButton::clicked, [=, &previewDlg]() {
```
atsju: there are 2 warnings related to this line that need to be fixed.
`captured local variable by reference might go out of scope before lambda is called [-Wclazy-lambda-in-connect]`
and
`Pass a context object as 3rd connect parameter [-Wclazy-connect-3arg-lambda]`

foucaultview.cpp
```diff
@@ -152,280 +169,351 @@ QVector<QPoint> scaleProfile(QPolygonF points, int width,
...
-
-
-                if (x> hx - ppl/2. && x < hx + ppl/2.)
-                {
-                    ronchiSlit[0].at<double>(y,x) = 1;
-                }
-
+    sv->setSurface(wf);
+    cv::Mat surf_fft = sv->computeStarTest(s.heightMultiply * sv->nulledSurface(effectiveZ3), size, actualPad, true);
+    wf->InputZerns = originalZerns; // Restore state immediately
```
atsju: This is a copy of https://github.com/githubdoe/DFTFringe/pull/309/changes#r2647171212
I just put it again here so that we don't have to dig to find the comment.

When I have some time I will try to find if it's realy brocken in some way the user can reproduce.

The pattern I see is : 
```C++
wf->InputZerns = tempZerns;
sv->setSurface(wf);
wf->InputZerns = originalZerns; // Restore state immediately
```

at first look this seems quite dangerous. Code is passing `wf` pointers and storing these pointers. The fact that there are temporary zernike will be lost as soon as the code is over because there is no actual copy and code works on same object. 
Maybe it just works because of the order of display calls but I'm affraid it will be very difficult to maintain and bugprone for example changing an unrelated display setting could redraw with wrong data ?

Please check and advice

atsju:
> Concerning the 550.
> 
> There is now a new routine that users can call with this structure that makes ronchi images. The old code that made ronchi images has been refactored into using that new make ronchi image code. That new routine uses this structure with the 550 in it.
> 
> The caller can use those default values if desired however the refactored code fills them in and in the case of the 550 it fills that in from the preferences data. Go look at the code that references that structure and you will see. Using Qt Creator it is easy to find all code that references something using a right click on the highlighted something.
> 
> One way that structure could have been written is with no default values. Then there would be no 550. However AI elected to fill it with reasonable default values. If you think you need it put a comment out beside the 550 that describes it as the output wave length in nanometers.

OK it's clearer. I had a look and actually all the 550 from these modifications are never loaded from settings as you think. 
QtCreator probably mixed up `OpticalTestSettings.outputLambda` from the structure and the variable with same name `double outputLambda;` defined in surfacemanager.cpp
The whole usage of `outputLambda` is quite convoluted as you need to trigger ui updates when value is changed.

So yes I think this code is wrong in some places. While it's correct to do `double outputLambda = 550.0;` in foucaultView.h line 33,
it's not OK in fouclaultview.cpp lines 294 and 362. For the latest, it shall load from settings.

Let me know if I missed something

foucaultview.cpp
atsju: `[param, param, param](){code using param}`
I wasn't that familiar. It's called a lambda function if you want to search.
Basically = means all parameters in the context are copied and usable. There are other syntax possible than =.

I added context to the `connect` because parameters where copied but the dialog parameter passed by pointer would not exist anymore after function execution. Thus you don't want to call that connect once function is over or it will fail. And this is why there was a warning. Adding context gives automatic disconnect and fix the warning.
There was no actual problem here as exec is blocking and the connect would never get called outside the function. Warning was still a good heads up.

githubdoe:
output_lambda good catch.  The original code actual uses an external variable that is set by other things and was the awkward way I got that value to foucaultview.  Yes the current code is wrong.

foucaultview.cpp
gr5: sounds good.

atsju: hi @gr5 you resolved the conversation. The change above needs to be commited first (also your 2 other suggestions about askAboutInversion).

gr5:
I think it's fine to merge this now.  Should I go ahead and just click the button?  Or does someone else want to add anything first?  I think this is ready.  It's not thoroughly tested by me but I'll test it more thoroughly when we do a release.

PR #310, opened 2025-12-24
Made the 2 docking splitters have slightly wider, teal colored, much …

gr5:
I simply made these draggable partitions more visible to hopefully make it easier for users to realize that they can resize the left and right docks.

image

I'll post how it looks at 200% scale once this PR creates the QT6 build (I have 2 windows computers and I'm at our second home on my laptop which doesn't have QT6 configured yet)


gr5:
Here it is at 200% (built by QT6).  I think it would be fine if I had a 4k display instead of just 2560.

image

gr5:
Please review.

PR #316, opened 2026-01-07
Fixes crash in "percent of correction" feature

gr5:
So this may be crazy because I might be the only person on the planet with these bad registry values.  I don't know how I created these registry values but it was just me messing around and testing the feature a while back (maybe a few months?).  It's possible that the only version of DFTF that can create these bad registry values was never released but just in case I want this fixed.

Anyway these were my values for 2 items in the registry and they caused a crash:

```
correctionZones              @ByteArray({"ROC":0,"mirror radius":73.025,"zones":[]})
percent number of zones      0
```
The code changes in this PR forces DFTF to go to default zones (5 zones with reasonable values) if zones is zero.

### DIFF
--- a/RevisionHistory.html
+++ b/RevisionHistory.html
@@ -10,7 +10,7 @@
     <h1>DFTFringe Version History</h1>
     <ul>
 
-        <li>Version 8.3.0</li>
+        <li>Version 8.3.1</li>
         <ul>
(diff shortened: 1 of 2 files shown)


githubdoe:
Sounds good.  Perhaps also  limit the spin button to 5 as the low limit.

atsju (approved):

gr5:
I'm done testing 8.3.0 and I guess I'm not going to wait for Dale to answer my question about the spin button limit so I'm going to merge and do some minor testing and release 8.3.1.

PR #321, opened 2026-01-08
Allows compatibility with older wavefront (.wft) files

gr5:
Sorin noticed this "bug" introduced in v7.3.1. This only affects people who create wft files with other software.  Software other than DFTFringe.


atsju (approved):

githubdoe:
I think it should be put to what it use to be.  I'm not really sure if i changed it now.  But if I did it was for debugging purposes and forgot to put it back.

Issue #320, opened 2026-01-08
ronchi/foucault simulation lost ability to do autocollimation simulation


gr5:
This is a new bug introduced in 8.3.1 (and 8.3.0).  Commit was this commit:
d4e87ae1efdaaf0b56d4c6829aa53b560b93cb22
Made 12/22/2025 at 20:54pm mountain time by Dale

I believe Dale said that this change was made by AI and it looks like it because the comments are not Dale's style.  The function on_makePB_clicked() was completely rewritten and in the old version we had this code:

```
    bool oldDoNull = md->doNull;
    if (!ui->autocollimation->isChecked()){
        md->doNull = false;
    }
```
And then later doNull is restored to the old state.

It may be as simple as adding this back in.  I'm not sure quite where to put it because the code has changed a lot.  hopefully no other features got lost.

doNull seems to be used in a few functions but I'm guessing it's computeWaveFrontFromZernikes() that is important for ronchi/foucault.  And that hasn't changed.  So I think this is not too hard to fix.

PR #322, opened 2026-01-08
Fixed bug in foucault/ronchi view where it was always in AC mode

gr5:
In the end it was a simple fix.  The foucault/ronchi creation was moved into a static function so I passed a new argument to that function whether or not it should be in autocollimation mode.

I tested for the primary display and also the multi-ronchi view.

### DIFF
--- a/RevisionHistory.html
+++ b/RevisionHistory.html
@@ -12,7 +12,7 @@
 
         <li>Version 8.3.2</li>
         <ul>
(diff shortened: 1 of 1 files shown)


atsju:
> Fixes #320

Thus does not work. If you put same message in top comment, it will autoclose :)

atsju (approved):

PR #324, opened 2026-01-08
average profiles now get slope feature as well where slope is in orange.

gr5:
I don't understand what I did but it seems to work.  I don't understand the calculations of xDel and hDelLimit and so on.  Well I get lambda - that's the wavelength.  But it seems to work perfectly so I think I did it right.

All the code is cut and pasted except as you can see I changed an object from class QwtPlotCurve to ProfileCurve (ProfileCurve has ability to show slope and QwtPlotCurve does not).

Anyway I tested it against the non average and it seems to show slope in the same steepness.

### DIFF
--- a/profileplot.cpp
+++ b/profileplot.cpp
@@ -820,7 +820,17 @@ qDebug() << "Populate";
               left = transform.map(left);
               left.append(right);
 
-              QwtPlotCurve *cprofileavg = new QwtPlotCurve( name + " avg");
+              ProfileCurve *cprofileavg = new ProfileCurve( name + " avg");
+              if (left.size() >= 2) {
+                  // Distance between two samples
+                  double xDel = fabs(left[0].x() - left[1].x());
+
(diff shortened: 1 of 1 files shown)


githubdoe:
Yes the new profile curve class knows how to plot the slope error.

xDel is the x dimension over which the slope will be calculated.  hDelLimit is how high the y can vary before the slope exceeds the limit set by the user.

gr5:
> I don't know why you added the code to set the slope settings. I would think they were already set.

No I had to set the slope settings.  Because just above is that "new" which created an instance of class ProfileCurve and so I had to set the boolean m_showSlopeError with setSlopeSettings().  The boolean isn't a static.  It's a normal member variable.

If you look at the code in entirety there is a section that creates the single profiles, the 16 profiles and the average profiles and each section creates their own instances and each section of code needs to set the slope flag thing.

PR #331, opened 2026-01-15
added blink compare to ronchi compare

githubdoe:
New code to blink top two ronchi selected wave fronts


githubdoe:
All objects created by a QDialog that have it as their parent are deleted when the dialog is deleted.   Widgets that are in a layout have the layout as their parent and the layout eventually has has the dialog as it's parent.

PR #351, opened 2026-08-03
Documentary : improve crashlog .dmp file usage documentation

atsju:
I have had some nasty crash due to "remove lens distorsion" (bad usage when testing the UI) that I wanted to investigate.
This method helped me troubleshoot it. It's only documentation update.

# Technical explanation :
DFTFringe is built with MinGW/GCC using force_debug_info and separate_debug_info, which produces a DFTFringe.exe.debug file containing DWARF debug information. Crash dumps are standard Windows minidumps generated with MiniDumpWriteDump().

WinDbg can open the dump and provide a valid stack trace, but because the application is built with MinGW, it cannot fully resolve symbols and source locations like it would with a PDB-based build (MSVC or Clang/MSVC). Application frames therefore usually appear as DFTFringe+offset.

To obtain the corresponding source file and line number, use addr2line with DFTFringe.exe.debug. The offset reported by WinDbg is an RVA and must first be converted to the image virtual address expected by DWARF by adding the executable image base (0x140000000).

In short:

MSVC and Clang/MSVC generate PDB files, which are fully supported by WinDbg.
MinGW generates DWARF debug information stored in DFTFringe.exe.debug.
WinDbg can analyze the minidump but cannot fully use the DWARF symbols.
Use WinDbg to identify the crashing address, then use addr2line on DFTFringe.exe.debug to recover the function, file and line number.

We might want to build with a different compiler then mingw in the futur.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,7 +314,25 @@ aqt install-tool windows desktop tools_ifw
 
 :warning: This method is for Windows only.
 
+## If you get a stack trace in .log file
+
 - Get `DFTFringe.exe.debug` file corresponding to the release where the log comes from in github
-- Run `addr2line -e DFTFringe.exe.debug 0x00000000005A7229` with the address of the stack you get from the log
+- Run `addr2line -e Z_DFTFringe.exe.debug 0x00000000005A7229` with the address of the stack you get from the log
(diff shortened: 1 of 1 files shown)


gr5 (approved):

gr5:
This is great.  I could have used this information in the past.

atsju:
> This is great. I could have used this information in the past.

Me too. But I havent been able to exploit it in the past. Experience ... and AI.

PR #352, opened 2026-08-03
fix outline helper not closing on app close

atsju:
Fix one of the windows not closing identified in #82 

--------
This PR modernizes Outline Helper lifecycle management by replacing its static singleton access with a MainWindow-owned dialog instance.

The intent is to make window behavior more predictable: the dialog now follows Qt parent-child ownership, so it closes naturally with the main window.
It also improves separation of concerns by having the dialog emit user-intent signals (Start/Stop), while MainWindow handles the actual processing logic. This keeps UI code focused on interaction and makes behavior easier to reason about, test, and maintain over time.

### DIFF
--- a/jitteroutlinedlg.cpp
+++ b/jitteroutlinedlg.cpp
@@ -1,23 +1,13 @@
 #include "jitteroutlinedlg.h"
 #include "ui_jitteroutlinedlg.h"
-#include "mainwindow.h"
-jitterOutlineDlg *jitterOutlineDlg::m_instance = 0;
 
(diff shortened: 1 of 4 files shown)


gr5:
I looked at the code changes.  It looks good.

gr5 (approved):
Code looks good.  I built and ran this with QT5 (of course I will test the QT6 build on next release). It closes the jitter window (outline helper) when you close the app.

atsju:
Thank you for your test @gr5. I hadn't tested it yet. Hence the "draft" pull request.
Please merge whenerver it's Ok or after Dale reviews.

Issue #353, opened 2026-08-03
[MacOS] spdlog_ex: Failed opening file DFTFringeLogs/log.txt for writing


Lucassifoni:
When launched **graphically**, DFTFringe instantly crashes on Mac. This is because the Finder (MacOS's file explorer and app launcher) runs applications with CWD = "/", which is not user-writable. The app then tries to create DFTFringeLogs/log.txt at / and crashes.

Running it from a terminal from ~Desktop (user-writable) : 

```
 Desktop ./DFTFringe.app/Contents/MacOS/DFTFringe
[2026-08-03 17:46:48.089] [info]


-------------
[2026-08-03 17:46:48.089] [info] spdlog info
[2026-08-03 17:46:48.089] [warning] spdlog warn
[2026-08-03 17:46:48.089] [error] spdlog error
[2026-08-03 17:46:48.089] [critical] spdlog critical
[2026-08-03 17:46:48.090] [critical] This is a demo stacktrace:
 0# main in /Users/lucas/Desktop/DFTFringe.app/Contents/MacOS/DFTFringe

[2026-08-03 17:46:48.198] [info] DFTFringe 9b8ffad25efb8f3c11f82b8b9103bd026f157edf started
[2026-08-03 17:46:48.606] [trace] Log level changed to: Info
```

Running it from / :

```
➜  / ./Users/lucas/Desktop/DFTFringe.app/Contents/MacOS/DFTFringe
libc++abi: terminating due to uncaught exception of type spdlog::spdlog_ex: Failed opening file DFTFringeLogs/log.txt for writing: Operation timed out
[1]    37407 abort      ./Users/lucas/Desktop/DFTFringe.app/Contents/MacOS/DFTFringe
```

I'm preparing a PR for this one.

atsju:
Closing manually as done.

I just discovered a new feature : https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-auto-closing-issues#about-auto-closing-issues
in the past, you needed to write keyword "close #<issue>" or "fix #<issue>" for it to work

PR #354, opened 2026-08-03
Adds Mac OS branch for the rotating log file opened at startup

Lucassifoni:
Related to https://github.com/githubdoe/DFTFringe/issues/353

I think we could remove the branching and standardize on getting a writable directory through `QStandardPaths::writableLocation`, what do you think ?

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -25,6 +25,7 @@
 #include "messagereceiver.h"
 #include "utils.h"
 #include <QDebug>
+#include <QStandardPaths>
 
 #ifdef _WIN32
     #include <Windows.h>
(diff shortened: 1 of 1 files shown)


gr5:
I prefer the way you did it where it only uses that function for mac.  But @atsju  could probably convince me otherwise.  He wrote that code originally.

I approve this change but want @atsju to look at this before merging.

atsju (approved):
I agree with @gr5 I prefer it this way too. Single location is easyer to understand by end users even if it's old school.

On windows, the logs would end in `C:/Users/<USER>/AppData/Roaming/<APPNAME>`. A better place would probably be `HomeLocation` but I dislike when applications do that too.

Lucassifoni:
I understand how the generic writable directory handed by the OS can be annoying for users who want to actually access the log. I'm fine with keeping the conditional then.

atsju:
As this change does not affect any functionality and has no risk on windows I'm merging it.

PR #355, opened 2026-08-04
fix #345 where 16 diameters view change base angle on redraw

### LINKED ISSUE #345: 16 Diameters View Randomizes Start Point
**Describe the bug**
When I have "show 16 diameters" checked, click on a wavefront from the selection menu, and press enter repeatedly, the wavefronts seem to shift. My theory is that the starting angle is not properly loaded from the angle-setting clock face widget when you load a wavefront from the wavefront menu, but instead only set when you interact with the angle-setting clock face widget.
It would be nice if the 16 chosen diameters were consistent and started from the direction specified in the little compass wheel.

**To Reproduce**
1. Load a wavefront from an astigmatic mirror
2. On the results screen, click "show 16 diameters"
3. From the list of wavefronts menu, click on your wavefront
4. Press enter repeatedly. (Or click on the wavefront from the menu repeatedly)

** Expected behavior**
The profile view remains the same.

atsju:
fix #345

### DIFF
--- a/profileplot.cpp
+++ b/profileplot.cpp
@@ -350,6 +350,7 @@ void ProfilePlot::setDefocusValue(double val){
 QPolygonF ProfilePlot::createAverageProfile(double /*umnits*/, wavefront * /*wf*/, bool /*removeNull = false*/){
     surfaceAnalysisTools *saTools = surfaceAnalysisTools::get_Instance();
     QList<int> list = saTools->SelectedWaveFronts();
+    const double baselineAngle = g_angle;
     QPolygonF avg;
     for (int indx = 0; indx < list.size(); ++indx){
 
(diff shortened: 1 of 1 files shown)


gr5:
You fixed it but I think a related bug maybe?:

The contour graph has a dashed line showing the angle
When you click on the wavefront the dashed line goes away and doesn't come back unless you use the "compass" control to rotate it.

gr5 (approved):

atsju:
This makes the dashed line always visible from the start. No more discrepency when selecting different wft. Let me know if this is what we want.

gr5 (approved):
Works great.  I couldn't download the artifact so I just built it and it works nicely.

PR #359, opened 2026-08-05
Fix potential dip in profile plot when using Gaussian blur + minor display issues

### LINKED ISSUE #317: Very minor bug with contour settings tab
Bug is in ver 8.3.0 and probably many older versions.

If you have a contour plot visible, and choose the contour tab you can change the max or min values in the error scale area and the waves result gives you the difference between them.

However if you hover the mouse over the profile plot and use a scroll wheel to zoom in and out the min and max seem to correctly update but not the "waves" datum.

### LINKED ISSUE #319: Very minor gui issue - save file dialog says foucault instead of ronchi when saving ronchi
version 8.3.0

When you right click on ronchi or foucault view you can save the diagram.  If you choose either the save file dialog has this title as shown but for the ronchi the circled text should say ronchi.  Instead it says foucault.

Image

### LINKED ISSUE #333: weird dip in profile display
seen in v8.3.2 but I think all versions probably do this.

Minor bug.

Dale had an extremely similar issue (that I closed) that was in an older version but that got fixed I think but I think this issue is different.  Maybe.

If you have a wavefront that is an odd number of pixels (601 pixels for sure works for the bug) then the profile, when perfectly horizontal (at 0 degrees angle), will have one zero value at the end and **the profile sharply dips up or down on the last pixel.**  

It only happens at 0 rotation where the above mentioned for loop is getting a floating point off-by-one at the end of the loop.  

Image

atsju:
fix #333 

Problem was due to Gaussian blur not using same inclusive/exclusive boundaries as cicle mask code (second commit)   
I also changed some float to int truncature into rounding as small epsilon could cause issues with sin/cos (first commit).

Fix #319 
Fix #317

### DIFF
--- a/utils.cpp
+++ b/utils.cpp
@@ -227,9 +227,10 @@ void CropGaussianBlur(const cv::Mat &in_, cv::Mat &out, int kernelSize, const Ci
     double radius2 = (outside.m_radius-2)*(outside.m_radius-2);
     for (int x=0; x<width; x++) {
         double dx = (cx-x)*(cx-x);
-        for (int y=0; y<width; y++) {
+        for (int y=0; y<height; y++) {
             double dy = (cy-y)*(cy-y);
-            if ((dx+dy) >= radius2) {
+            // Keep boundary pixels inside, consistent with fillCircle() mask creation.
+            if ((dx+dy) > radius2) {
                 mask.at<double>(y,x) = 1e-5; // can't use zero or we get divide by zero issues later (in cells we don't care about)
                 in.at<double>(y,x) = 0;
             }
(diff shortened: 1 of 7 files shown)


gr5:
Something wrong with github - these "download installer here" links don't work and if I go to actions to get them from there - also not working.  Instead I have to build these myself.  No big deal but I've heard that github has lots of new bugs related to AI somehow?

atsju:
Very strange. I can download the installers (I did not even build once on my own computer this week).
I didn't hear of new bugs but sure GitHub business model is changing for enterprise users.

atsju:
It's probably easier you just wait for it to work again.
Eventually you can check if you can download artifacts form a pull request openned by yourself. And I will try your artifact. This would cross check if it's an issue with rights/privilege management.

utils.cpp
atsju, line 208: https://github.com/githubdoe/DFTFringe/blob/7b78459d3afd55863ce1014c18aa64826867ff43/surfacemanager.cpp#L638
https://github.com/githubdoe/DFTFringe/blob/7b78459d3afd55863ce1014c18aa64826867ff43/surfacemanager.cpp#L641

From my understanding the arbitrary shrink comes from there and this is the "other place" to mess with.

gr5: Yes.  Probably.  It's tough to know if one caught all the spots.

gr5 (approved):
Please don't merge if we are doing a release where I don't have a chance to test every feature in DFTF.

I tested the hell out of this feature (odd and even diameter wavefronts, averaging wavefronts, subtracting, ronchi, foucault, saving and loading wavefronts) and it looks fantastic.

But it's a little crazy how changing one feature can affect a feature that is seemingly unrelated so I'd like to test everything else e.g. "test stand astig removal" on the next release with this code (again - it shouldn't be affected but I've gotten burned before).

gr5:
To be clear, I recommend we merge unless we are doing a release in the next 8 days.

Issue #152, opened 2024-05-20
annulus config is not recomputed


atsju:
**To Reproduce**
1. configure a mirror with annulus
2. import some pictures, outline and all the usual
3. "oh I forgot to change the config from previous mirror. let's remove annulus"
4. configuration-> mirror-> uncheck annulus. OK
5. Current wavefront is updated. Cool :)
6. look at next wavefront => It still shows annulus hole
7. open and close config window is enough to recompute the wavefront but it shall not be necessary.

**Expected behavior**
I expect that all wavefronts are recomputed when I display them.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: win10
 - DFTFringe version: 6.2.1
 - Build: official GH release

atsju:
I see the same for "Gaussian Blur diameter". I see it checked or unchecked but wavefronts are not updated on loading them. 
When any configuration common to all wavefronts is updated they should be all systematically recomputed. Selecting them all fixes the issue but is long to load. Maybe they should all have an associated boolean telling that they need or not recomputation. Once they have been loaded once, no need to recompute unless a config has changed.
OR they get all recomputed each time we display them but this will cause slowness.

gr5:
Same thing for checking and unchecking zernike's and then hitting "recalculate".

All these changes only affect the currently selected wavefront.  But when you switch to another wavefront, the checkboxes and values are reflecting the wavefront you were previously on.

To have these changes recalculate all the wavefronts and if there are say 30 wavefronts open - this might be very annoying to the user and take several minutes.

I'm not sure what the best solution is but I feel we need to figure out the best UX (user experience) first.  Probably when you switch wavefronts it sets all the gui elements back to what they are for that particular wavefront (in the mirror config as well though?  when you change the mirror config should there be an "apply to all wavefronts" button as well?)

The way it works now isn't too horrible once you realize it only affects the current wavefront until you hit "recalculate" button.  We could also have a different color for wavefronts that aren't matching the gui and/or a recalculate button next to each wavefront that hasn't been updated.  And a recalculate all wavefronts button in the same region of the GUI.

githubdoe:
The current behavior is on purpose.  That is only the selected wave fronts should be affected.  Yes sometimes you want them all so then you have to remember to select them all.  There are many times I don't want them all so this works for that case.

githubdoe:
I should add that is the behavior I want because I quite often want to compare wave fronts with different settings.

atsju:
[...]

PR #361, opened 2026-08-05
fix gaussian blur checkbox not displaying actually applied settings

atsju:
This improves #152 

Now each wft remembers if gaussian blur is enabled or not. I implemented as requested by Dale discussion he want's to be able to have different settings in different wft.

I'm working on other settings in different PR


gr5 (approved):
Works great.  Code looks good.  Tested. Approved.

PR #364, opened 2026-08-06
fix elliptical flat configuration bug and ensure minor axis is always smaller

### LINKED ISSUE #301: bug with mirror config regarding "elliptical flat" option
To test an elliptical igram you check a box in mirror configuation ("Elliptical flat...").  But the text to the right of this is never enabled.  This is a super easy fix.

This bug was introduced in version 5.1 back in 2021.  More specifically the commit after the commit tagged "5.0" and with the commit comment "Revision5.0".

Anyway the fix is approximately 2 lines of code that would enable/disable this text field based on the value of the checkbox (more than one line of code as you have to worry about this when you load a file as well).

**WORKAROUND**

For version 5.1 through 8.2 there is a workaround.  First uncheck "elliptical" checkbox.  Then enter mirror diameter to the value that appears in the box to the right of "vertical axis".  Then check the box and enter what you want vertical axis to be into the diameter field.  This will change both fields.  Then uncheck the box, enter what you want diameter to be and then check the box again.  Now you have the correct desired values in for both fields.  Now you can draw the outline and it will be the correct ratio/shape.

atsju:
fix #301

I did not test this one beyond "mirror configuration" responsiveness.

### DIFF
--- a/mirrordlg.cpp
+++ b/mirrordlg.cpp
@@ -116,6 +116,7 @@ mirrorDlg::mirrorDlg(QWidget *parent) :
     ui->ellipseShape->setChecked(m_outlineShape == ELLIPSE);
     setEllipseControlsEnabled(m_outlineShape == ELLIPSE);
     ui->minorAxisEdit->setText(QString().number(m_verticalAxis));
+    enforceEllipseMajorAxis();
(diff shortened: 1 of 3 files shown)


gr5 (approved):
Code looks good.  I tested this as well.

PR #365, opened 2026-08-06
Provide raw Linux and AppImage builds as part of CI

Lucassifoni:
Hello,
The idea is to provide both bare compilation outputs (should run under most *buntu flavors provided you have qt installed) and an AppImage build (assembled under 22.04 only for wider compatibility) as part of CI. This is quite cheap since the builds were already done.

@atsju I looked into QT Installer Framework, but from my understanding it wouldn't make the overall build workflow shorter. It could be used to integrate more tightly with desktop OSes (provide menus, icons, etc) but this would be another scope. 

Another great next step would be to actually launch DFTFringe in a desktop *buntu container to catch eventual regressions on Linux since there aren't that many users. I will see if that's doable next week.


atsju:
GitHub is experiencing worldwide issues with action. I just closed and reopened to trigger the pipelines.

.github/workflows/build-linux.yml
```diff
@@ -23,12 +23,30 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - run: sudo apt update
-      - run: sudo apt install -y apt-utils build-essential wget qt6-base-dev-tools qt6-declarative-dev qt6-multimedia-dev libqt6charts6-dev libqt6datavisualization6-dev libqt6svg6-dev libopencv-core-dev libopencv-dev libqwt-qt5-6 libqwt-qt5-dev libarmadillo-dev libgl1-mesa-dev libglu1-mesa-dev bear
+      # imagemagick/file/libfuse2 are only needed for the AppImage steps but are
+      # cheap to install everywhere. libqwt-qt5-* dropped: Qwt is built from
+      # source below and the distro Qt5 flavor was never linked against.
```
atsju, line 28: thank you for this change comment. Should probably be removed in source

.github/workflows/build-linux.yml
```diff
@@ -23,12 +23,30 @@ jobs:
...
+      - name: Put release name into a variable (tag)
+        if: ${{ startsWith(github.event.ref, 'refs/tags/v') }}
+        run: echo "WORKFLOW_VERSION=${{github.ref_name}}" >> $GITHUB_ENV
+      - name: Put release name into a variable (single commit)
+        if: ${{ !startsWith(github.event.ref, 'refs/tags/v') && github.event_name != 'pull_request' }}
+        run: echo "WORKFLOW_VERSION=${{github.sha}}" >> $GITHUB_ENV
+      - name: Find and Replace MY_AUTOMATED_VERSION_STRING
+        run: sed -i "s/MY_AUTOMATED_VERSION_STRING/${{env.WORKFLOW_VERSION}}/" DFTFringe.pro
+
+      # ---- build (unchanged) ----
```
atsju:
```suggestion
      # ---- build ----
```

.github/workflows/build-linux.yml
```diff
@@ -50,3 +68,74 @@ jobs:
...
+      # Note: the bare binary expects Qwt in /usr/local/qwt-6.3.0 (rpath) and the
+      # distro Qt6/OpenCV/Armadillo packages on the target machine.
+      - name: Pack bare build
+        run: tar -czf DFTFringe-bare-${{ matrix.os }}.tar.gz DFTFringe
+      - name: Upload bare build
+        uses: actions/upload-artifact@v7
+        with:
+          name: DFTFringe-bare-${{ matrix.os }}
+          path: DFTFringe-bare-${{ matrix.os }}.tar.gz
+          if-no-files-found: error
```
atsju, line 83: It's unclear to me, why do we upload the bare build ?  I though we require mainly the appImage for end user

atsju (commented):
Nice work.
could you please update the Documentation with Linux Appimage usage ? Then we have both compile and usage documentation for all 3 platforms.


Also it would be great to have the "Comment PR with artifact link" comment all 3 binaries (Win, MacOS, Linux) at once when all build success. It should probably be implemented in a separate PR.

.github/workflows/build-linux.yml
Lucassifoni: I should have explained it better as part of the initial PR. Some linux users (me included) prefer bare builds, but with your comment I'm starting to think that kind of user can just compile themselves. Let's focus on AppImages

atsju: From my understanding this is too OS-specific to distribute. I agree those people can build easily themselfs.

atsju:
I also forgot to mention: `make-release.yml` needs to be modified to publish the appImage

Lucassifoni:
I think this is ready too, it runs fine on various *buntu derivatives.

atsju (approved):
Code looks good. 
Generated .appimage runs in my WSL ubuntu. Everything is fine

~~Note: build-linux actions are hanging right now. I believe there is a problem with some apt mirrors. I suppose it will fix itself in coming hours.~~ It solved itself as expected

githubdoe (approved):

gr5:
You don't need my approval at this point, right?  I can't wait until it's working so I can try out the AppImage.

atsju:
@gr5 you can already try the app image. It's in the build artifacts. I can link you if you need.

We can probably modify the "new build artifact" to have all 3 OS in same comment. Can do in a futur PR. 
As for now, the artifacts will show automatically only on releases

atsju:
Bottom of page here https://github.com/githubdoe/DFTFringe/actions/runs/32247235250

gr5:
I looked through the build artficats but couldn't find it.  Thanks.  It seems to work just fine in Ubuntu.

PR #369, opened 2026-08-11
zernike enable table is saved for each wft. Display diff

atsju:
This one also improves #152 . We can probably do it differently and with better visuals but it's enough to validate the intent. 
Just let me know.

you will notice small  + and - before the zernike names. 
It makes it much clearer witch zernike have been used to display a given waveform and what would change if hitting compute button.

### DIFF
--- a/surfacemanager.cpp
+++ b/surfacemanager.cpp
@@ -470,10 +470,14 @@ void SurfaceManager::generateSurfacefromWavefront(wavefront * wf){
     wf->nulledData.release();
 }
 cv::Mat SurfaceManager::computeWaveFrontFromZernikes(int wx, int wy, std::vector<double> &zerns, QVector<int> zernsToUse){
-    double rad = getCurrent()->m_outside.m_radius;
-    double xcen = (wx-1)/2, ycen = (wy-1)/2;
-
+    wavefront *currentWf = getCurrent();
     cv::Mat result = cv::Mat::zeros(wy,wx, numType);
+    if (currentWf == nullptr) {
+        return result;
+    }
+
+    double rad = currentWf->m_outside.m_radius;
+    double xcen = (wx-1)/2, ycen = (wy-1)/2;
 
     double rho;
 
@@ -482,9 +486,7 @@ cv::Mat SurfaceManager::computeWaveFrontFromZernikes(int wx, int wy, std::vector
         if (value > maxZernToUse)
             maxZernToUse = value;
     }
-
-
-    std::vector<bool> &en = (getCurrent()->zernEnablesApplied.empty()) ? zernEnables : getCurrent()->zernEnablesApplied;
+    std::vector<bool> &en = (currentWf->zernEnablesApplied.empty()) ? zernEnables : currentWf->zernEnablesApplied;
     mirrorDlg *md = mirrorDlg::get_Instance();
     for (int i = 0; i <  wx; ++i)
     {
--- a/surfacemanager.cpp
+++ b/surfacemanager.cpp
@@ -931,7 +931,22 @@ void SurfaceManager::computeMetrics(wavefront *wf){
     wf->min = mmin * md->lambda/outputLambda;
     wf->max = mmax * md->lambda/outputLambda;
 
-    ((MainWindow*)parent())->zernTablemodel->setAppliedEnables(&wf->zernEnablesApplied);
+    // let zernTableModel know state of currently selected wavefronts zernike state (which were checked/unchecked/both)
+    QList<int> doThese =  m_surfaceTools->SelectedWaveFronts();
+    bool bFirst=true;
+    foreach(int ndx , doThese){
+        wavefront * each_wf = m_wavefronts[ndx];
+        if (bFirst) {
+            ((MainWindow*)parent())->zernTablemodel->setAppliedEnables(&each_wf->zernEnablesApplied);
+            bFirst = false;
+        } else {
+            ((MainWindow*)parent())->zernTablemodel->addAppliedEnables(&each_wf->zernEnablesApplied);
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


gr5:
I love this feature!  Chef's kiss.

It even keeps track if I select multiple wavefronts, change a zernike checkbox and hit recompute.  It properly keeps track of which wavefronts were modified and which weren't.

There is a minor feature/bug however:

If I have 2 wavefronts, one was computed with spherical zernike checked and one without.  I can click on the two wavefronts one at a time and I can see which one was calculated with and without.  however if I select both wavefronts then it matters which I select last.  Should there be a third symbol when it's mixed?

With software installers there are checkboxes that can be checked, unchecked, or there is a 3rd symbol to indicate that some of the sub features are checked and some are not.  I don't want to change the checkboxes here but the "+" and "-" symbols perhaps to a third symbol "." or "*".

gr5:
Are you going to try to add a third symbol @atsju?

atsju:
yes it sounds like an excellent idea.
I need some time.

more generally, I had in mind to fix lot of little identified issues but at low pace. Just let me know if you have release in mind.

githubdoe (approved):

gr5:
I wanted to release this week but now Dale has a new feature and he hasn't done a PR and he may be adding more features so no rush at the moment.

githubdoe:
I'm playing and do not have anything to release.  When it is ready I want others to test it and decide if it is worth while and the GUI makes sense.  So it is a long way away from being a released feature.

gr5:
I thought we could release it as an "experiment" so we can play with it and think about how it should be done "right" and also see what video devices it works with and what ones it doesn't.

Changing the subject a little: because I was testing out doors I found it very useful to average 10-20 wavefronts without moving anything just to get rid of the air current noise.  The wavefront would really settle down even with gusts of wind.  Without the averaging it was a mess - bouncing around.

This feature might only be useful when you are out doors or if you have a fan running.  I don't know.  But at Stellafane it was very useful.

So I kind of envision making 10 averages (one wavefront made from 10).  Saving one wavefront.  Moving tilt, get a new outline, make another 10-igram average.  Or alternatively after the first 10-igram average, rotating the mirror by 90 degrees and getting a second 10-igram average.

There is a guy, Charles Wright, who claims he has a rock solid, very useful, outline generator.  If we have that as well only then would we likely be able to move the tilt while it's averaging.

I don't think the "blue circle" needs to reach out to the lobes - I have tested a small blue circle and it barely changes the results.

But I'm okay with doing a release next week without the video feature.

githubdoe:
There is sort of a paradigm shift with the blue circle when using the live video.  You are use to setting the blue circle based on the DFT.  With live view we can set a blue circle and then change tilts and defocus to always make the sidelobes at least just outside it.    The blue circle helps get rid of low spatial frequency background noise.  

Currently when you start the loop it starts showing a running average that contains all the wave fronts since it started.  It saves each individual wave front in the wave front list as usual.  It does not save the average it just shows it.  You can wait watching till the average settles down.  Then stop and go over to the wave front list and select them and average them in the usual way.  Then you have a real average wave front with as many as you need.

githubdoe:
To test it you need either skysolve or a USB camera.  You can of course run analsis on any image the USB camera has in it.  It will create a surface of some sort.  I was surprised to see a surface kind of like the glasses on my face when using my web cam on my face.  Many waves of error but setting the y scale to about 100 wave create a flat surface of my glasses outline among a lot of other blotches.  
image

gr5:
I can do the * for when there are both (+ or -)  and  nochange.  If you want @atsju.  I should have plenty of time to do this Friday or Saturday.

atsju:
If I havent done it by end of the day (french time), feel free to do it. Sorry and thank you !

atsju:
Hi, I won't have time to do it before next week (potentially) sorry

gr5:
I made 2 changes!

1) I didn't add the [*] idea as it made more sense to stick with the [+] and [-] but if it's mixed then it shows one of those.  More details:
Let's say you just unchecked astig.  Now all the astig for all wavefronts have the [-].  If you recalculate just one wavefront it will now lose the [-] but the other's still have the [-].  If you select multiple wavefronts and it's mixed such that some have astig and some don't it will still show the [-] as it will indeed remove astig if you recalculate selected waverfronts - it's just that one already had it removed.

2) I changed your tooltips.  I didn't like how they were for the default - when there is no [-] or [+].  Here is a video showing the tooltip changes:


https://github.com/user-attachments/assets/e5b86403-3339-436f-b9d2-5b74f15556be

gr5:
@githubdoe or @atsju please approve my latest change (github isn't stopping me from merging but my change wasn't reviewed at all).  My code edits are here: https://github.com/githubdoe/DFTFringe/pull/369/changes/6d2fdf74f2fae21dd5d95726b9300b704baac5b4

Description of new behavior is in my most recent post above.

atsju:
Thank you George.

I can review code tomorrow evening or Monday only.

What about `[+*]` and `[-*]` on multiple selection with different settings? It doesn't feel necessary though.

gr5:
> What about [+*] and [-*] on multiple selection with different settings? It doesn't feel necessary though.

It would be easy to do -- now that the array contains 3 states (Enable/Disable/Mixed).  Just 2 more if statements.  But I don't really see a need and don't think it will be more clear.  And I don't think we should have a different tool tip for that either.

atsju:
Code style looks good, I did not find any obvious flaw. 
I regret not being able to undesrtand the code 100% in detail after 30 minutes but it works so it's good to go.

I also made github copilot review it with the free 200 credits. I can show you how to do it if you are interested.
It didn't find any major problem.

you can go for merge !

Issue #358, opened 2026-08-05
Importing wavefront with different config: "yes to all" doesn't work


atsju:
v8.4.0 when you import a wavefront with different settings

you need to hit "yes to all" 3 times.
<img width="498" height="132" alt="Image" src="https://github.com/user-attachments/assets/e1f80194-1237-4c13-a858-898ac2d36366" />

I was supposing it yes to all "settings". 
But it could also be interpreted as "yes to all wavefronts" for this specific setting.
By the way, Importing multiple differnt wavefront will trigger the dialog only once. This was only a quick test, needs to be investigated further.

gr5:
I think it's fine as it is.

atsju:
It might only happen in some configurations. I will push investigation more and share video example here.

gr5:
3 times.  It asks about wavelength, diameter, and RoC. 3 things. That's how it works.  If only 2 of those are changed it should only ask for 2 of them.  If those match then it doesn't have to ask.

atsju:
I had a look at the code. It's probably fine, but I find the current message slightly misleading.

At least, I think the text could be updated to:

> Do you want to make the **wavelength** config match ?

My initial understanding of the dialog was: "If I click "Yes to All", DFTFringe will make the config match all settings of this wavefront (wavelength, ROC, lambda)."

However, the actual behavior is: "If I click "Yes to All", DFTFringe will make wavelength match on all subsequent wavefronts." 

Consider the case where three wavefronts from three different mirrors are loaded. Which of these behaviors do we want?

- Silent wavelength matching for all mirrors after the first one.
- Prompt again for each new mirror, while still allowing all parameters of that mirror to be matched.

The first behavior is what we currently have. I had assumed we wanted the second, but you have more experience with the intended workflow. Please confirm, and we can close the issue after updating the dialog texts.

githubdoe:
The second is what we want.   Somehow that broke awhile back.  I think.

gr5:
Yikes, what?  I think the behavior does what it says it does.  So that's good at least.

In addition, I have a totally different issue:  I never used to understand if clicking "yes" was going to change the wavefront to work with current mirror settings, or if it was going to change the current mirror settings.  I understand now that it changes the mirror settings but that seems dangerous if you already have many wavefronts loaded with the old settings.

[...]

Issue #234, opened 2025-08-18
Improve preference user interface


atsju:
from https://github.com/githubdoe/DFTFringe/issues/224#issuecomment-3189637615

> I do think the preference user interface could stand improvement. It is rather clunky. Sometime apply is needed sometimes not. More pages are needed but it is hard to add them especially since they are widgets. I have wanted to add more preferences but have not since it is so hard. I welcome any rework to the preferences GUI.

Issue #224, opened 2025-08-14
refactor QSettings usage


atsju:
## Proposal for 8.0.0 release
I propose to create a `settingsManager` class that would abstract `QSettings`. 
Every place where we use `QSettings` we would instead use `settingsManager::getSomethingValue()` and other dedicated method to read/write each of the possible settings.

I would use this occasion to reorganize keys into a tree and rename keys for more clarity. Especially use same keys for json and QSettings.
I would also add documentation.

## Advantage

- Centralized listing of all QSettings.
- More maintainable in long term
- No risk of using wrong key or different default values in different places.
- Tress structure for easier parameter tracking in regedit

## Disadvantage

- Users will lose most of their settings moving to 8.0.0
- Each read/write will do a fresh instance of QSettings. Impact should not be too heavy.

## Code example: 
```C++
#ifndef SETTINGS_MANAGER_H
#define SETTINGS_MANAGER_H

#include <QString>
#include <QSettings>
#include <QVariant>

class SettingsManager {
public:
    // Delete constructor to prevent instantiation
    SettingsManager() = delete;

    // Example: Get a value of "something"
    static int getSomethingValue() {
        QSettings settings;
        return settings.value(somethingKey, defaultSomethingValue);
    }

    // Example: Set a value
    static void setSomethingValue(int value) {
        QSettings settings;
        settings.setValue(somethingKey, value);
    }

    // Add more get_/set_ methods as needed...
};

#endif // SETTINGS_MANAGER_H
```

What do you think ?

githubdoe:
If I as a user knew I would loose my settings I would never upgrade until a new feature that I thought was a must have emirged.

I have tone of .ini files in all my old mirror tests.  I would not like to have to go back to an old version (which I may have lost) just to look at those mirrors.  Yes I do go back quite often.

githubdoe:
If it ain't broke don't fix it.  Is my first rule.  Especially if it changes something the user will notice in a bad way.  If it is not fixing a bug or adding a user seen enhancement then it should not be considered as a high priority.  Perhaps not considered at all.  Yes I often miss spell a settings keyword when created a save and fetch pair.  Something I have learned to test more when creating and something that causes me more waisted time then I like.  I have learned to make sure I do a cut and paste of the keyword.

I don't like telling users about how to find keywords in the registry by the way.  It confuses some.  Yes there have been times that is the best work around.   But try to limit doing that.  It is not a user friendly place.  The registry.

[...]

PR #370, opened 2026-08-11
[not ready for review] fix mirror settings modified in registry before user press accept button

### LINKED ISSUE #121: Mirror dialog needs to revert when cancel pressed
when mirror dialog is visible and you change values (such as mirror diameter), there are two issues:

it changes it in the registry instantly
it changes it in the official variable visible to other code instantly
Dale may have fixed number 1 above already while working on annular polynomials (2/8/2024) but number 2 still needs fixing.

If we need to store variables like mirror diameter, even temporarily, they should go into a private variable until the Ok button is pressed and then copied to the location that all the other code accesses.

atsju:
fix #121
mentionning #224 #234

I'm really sorry this one is quite long. It was difficult to split and even now some ellipse related things are temporarly brocken.

---------------- 
# short version

The mirror dialog's settings were exposed as public members, allowing user interactions to corrupt the persistent copy even on Cancel. This PR implements a **draft pattern** + **settings facade** to fix it:

- `m_current` (persistent copy) is now protected; external code reads it via `currentSettings()` (read-only)
- `m_draft` (working copy) handles all user edits; discarded on Cancel, committed on OK
- New `SettingsFacade` provides single atomic save/load point; eliminates duplicate QSettings calls
- All 30+ UI handlers refactored to use `m_draft`; removed 13+ public data members

---------------- 
# Long version explanation

## Problem Fixed

The singleton mirror dialog's settings were exposed as public members scattered across the class, violating encapsulation and creating a critical flaw: the `Cancel` button could not reliably discard edits because the persistent settings copy was unprotected and could be corrupted by user interactions.

### 1. Architectural Changes

**New Infrastructure:**
- **SettingsFacade** (new singleton): Centralized access point for all application settings
  - Enforces friend-only access to store implementations
  - Single atomic save/load point for all settings persistence
  
- **MirrorSettingsStore & ApplicationSettingsStore** (new): Private store classes with friend-only access to `mirrordlg` and facade
  - Generated via X-macros for compile-time field validation
  - QSettings keys defined once in `settingsstores_fields.inc`
  
- **settingsstores_fields.inc** (new): Single source of truth for all settings schema
  - 18 mirror config fields (`diameter`, `roc`, `cc`, `flipH`, etc.)
  - 3 application fields (`projectPath`, `mirrorConfigFile`, `lastPath`)
  - Eliminates magic strings throughout codebase

**Dual-Copy Pattern in `mirrorDlg`:**
```
m_current  → persistent copy (last-saved state, read-only via currentSettings() to external code)
m_draft    → working copy for dialog edits (temporary, discarded on Cancel)
```

### 2. Transactional Edit Semantics

**Before Dialog Shown:**
- Constructor calls `loadDraftFromSettings()` for safety (in case dialog used without `showEvent()`)
- Both `m_current` and `m_draft` are initialized from persistent storage

**While User Edits:**
- All 30+ UI event handlers modify `m_draft` (working copy)
- External code still reads `m_current` via `currentSettings()` (persistent copy)
- User actions are completely isolated from other code

**On Dialog Close:**
- **OK/Accept**: Commits `m_draft → m_current → QSettings` atomically via facade
  - `m_current = m_draft` (merge working to persistent)
  - `SettingsFacade::instance().saveMirrorSettings(m_current)` (atomic save)
  
- **Cancel**: Discards `m_draft` entirely
  - Next time dialog shown, `loadDraftFromSettings()` reloads from persistent storage
  - User changes are completely reverted

**showEvent() Override:**
- Always calls `loadDraftFromSettings()` before dialog becomes visible
- Syncs UI with reloaded draft values
- Guarantees Cancel reverts to last-saved state (fixes issue #121)

### 3. Encapsulation Improvements

**Removed Public Data Members:**
- Old: `diameter`, `roc`, `obs`, `cc`, `flipH`, `lambda`, `doNull`, `fringeSpacing`, `aperatureReduction`, `m_useAnnular`, `m_outlineShape`, `m_verticalAxis`, etc. (13+ exposed members)
- New: Single private `MirrorSettings m_draft` struct

**Replaced With Managed Access:**
- External code: `mirrorDlg::get_Instance()->currentSettings()` returns `const MirrorSettings&` (read-only)
- Settings access is now type-safe, compile-checked, and change-tracked

**Added Inline Getters for Computed Values:**
```cpp
double getFNumber() const { return FNumber; }        // f-number (computed on-the-fly)
double getZ8() const { return z8; }                  // Z8 Zernike coefficient
double getMinorAxis() const;                         // Ellipse minor axis
bool isEllipse() const;                              // Outline shape query
bool shouldFlipH() const;                            // Horizontal flip query
const MirrorSettings& currentSettings() const;       // Full settings struct (persistent copy)
```

**Compile-Time Access Control:**
```cpp
class MirrorSettingsStore {
private:
    friend class SettingsFacade;      // Only facade can construct and call save()
    friend class mirrorDlg;           // Only dialog can commit settings
    // ...
};
```
→ No way for external code to call `save()` directly; enforced at compile-time by linker

### 4. Settings Persistence Flow

**Old Design:**
```
External Code → QSettings → mirrorDlg scattered members → QSettings
(Magic strings, duplicated keys, type-unsafe, vulnerable to corruption)
```

**New Design:**
```
External Code → mirrorDlg::currentSettings() (m_current) → SettingsFacade
                  ↓ (on dialog accept)
             SettingsFacade → saveMirrorSettings(m_current) → QSettings
                  ↓ (on dialog show)
             SettingsFacade → mirrorStore().load() → m_current + m_draft
```

## Benefits

| Benefit | Impact |
|---------|--------|
| **Data Integrity** | Singleton's persistent copy (`m_current`) is now protected from Cancel operations |
| **Backward Compatible** | No QSettings key changes—legacy config files load correctly |
| **Type Safety** | Struct fields replace magic strings; caught at compile-time, not runtime |
| **Testability** | Facade enables easy mock/override for testing without filesystem I/O |
| **Single Source of Truth** | External code reads `currentSettings()` (persistent), dialog edits `m_draft` (working) |
| **No Duplication** | Settings defined once in X-macros; auto-synced across load/save code |
| **Error Prevention** | Removed 13+ public data members; accidental modifications now impossible |

### Core Architecture
- **settingsfacade.h** / **settingsfacade.cpp** — New facade enforcing friend-based access control
- **settingsstores.h** / **settingsstores.cpp** — New stores with generated struct definitions
- **settingsstores_fields.inc** — X-macro field schema (shared with stores and UI)

### Mirror Dialog Refactoring
- **mirrordlg.h** — Replaced 13+ public members with dual-copy pattern + getters
- **mirrordlg.cpp** — Refactored all 30+ UI handlers to use `m_draft`, added `loadDraftFromSettings()` and `showEvent()`, simplified `on_buttonBox_accepted()`

### Build System
- **DFTFringe.pro**, **DFTFringe_QT5.pro**, **DFTFringe_Dale.pro** — Updated to include new settings files

---

#### 1. **Persist Unit Preference (mm vs. inches)** 
```cpp
// TODO: actually mm is not saved in settings. should probably be saved 
// as it's a user preference
bool mm;  // Unit display flag: true = mm, false = other units
```
**Current Behavior:** Unit preference (`mm`) is transient—resets to default on restart.

**Solution:** 
- Add `bool unitsMM` field to `settingsstores_fields.inc`
- Persist in `showEvent()` and `on_buttonBox_accepted()`
- Remove hardcoded `mm(true)` initialization

**Benefit:** Remembers user's preferred unit system across sessions.

---

#### 2. **Clarify & Consolidate Outline Shape API** 
```cpp
// TODO: to be fixed with #358. 
// Saving shape shall be asked as it is for ROC, lambda and diameter 
// and be saved using adoptWavefrontSettings
void setOutlineShape(outlineShape shape);
void setMinorAxis(double val);
```

**Current Problem:** 
- Two separate methods for modifying outline shape
- No guarantee that shape changes persist to `m_current` and QSettings
- Wavefront loading sometimes calls `setMinorAxis()`, sometimes direct member access

**Solution:**
- Rename `adoptWavefrontSettings()` to accept outline shape:
  ```cpp
  void adoptWavefrontSettings(double diameter, double roc, double lambda, 
                               outlineShape shape, double minorAxis);
  ```
- Remove `setOutlineShape()` separate methods
- Single call-point ensures all shape changes are atomic and persisted
- Wavefront loader makes ONE decision and commits all at once

**Benefit:** Eliminates inconsistent state where shape is changed but not saved.

---

#### 3. **Clean allispe outline helper** 
```cpp
// TODO: This is still not 100% clean
// One call from loading file should be integrated to adoptWavefrontSettings
// Other calls are outline helpers. Need to be clarified.
void setMinorAxis(double val);
```

**Current Confusion:**
- `setMinorAxis` is called from outlining window and can modify mirror setting at each outline.

**Solution:**
- To be investigated. Might be OK as is. Renaming to clarify could be enough

**Benefit:** Eliminates API confusion; clear roles for each method.

---

### Why X-Macros?
Reduces boilerplate and eliminates source-of-truth duplication. A single field definition in `settingsstores_fields.inc` automatically generates:
- Struct member declaration
- QSettings load code
- QSettings save code
- Type validation

Adding a new mirror setting requires exactly one edit location.

### Why Compile-Time Friend Enforcement?
Prevents accidental calls to `save()` from unintended code locations. The linker will reject any `save()` call outside the allowed friend scope—impossible to miss in code review or CI.

### Why Always Reload on showEvent()?
Ensures the dialog never operates on stale data if another dialog modified settings between invocations. Also handles programmatic dialog reuse without explicit reset calls.

---

## References

- Original Issue #121: "Cancel button doesn't revert mirror config"
- Issue #358: "Clarify outline shape persistence"

### DIFF
--- a/astigstatsdlg.cpp
+++ b/astigstatsdlg.cpp
@@ -1,5 +1,6 @@
 #include "astigstatsdlg.h"
 #include "ui_astigstatsdlg.h"
+#include "settingsfacade.h"
 #include "circleutils.h"
 #include "circle.h"
 #include <qwt_plot.h>
@@ -489,8 +490,7 @@ void astigStatsDlg::showItem(const QVariant &item, bool on, int /*ndx*/){
 
 void astigStatsDlg::on_zernikePB_pressed()
 {
-    QSettings set;
-    QString path = set.value("mirrorConfigFile").toString();
+    QString path = SettingsFacade::instance().appStore().load().mirrorConfigFile;
     QFile fn(path);
     QFileInfo info(fn.fileName());
     QString dd = info.dir().absolutePath();
--- a/averagewavefrontfilesdlg.cpp
+++ b/averagewavefrontfilesdlg.cpp
@@ -66,7 +66,7 @@ void averageWaveFrontFilesDlg::on_process_clicked()
             sm->generateSurfacefromWavefront(wf);
             cv::Scalar mean,std;
             cv::meanStdDev(wf->workData,mean,std,wf->workMask);
-            double stdVal = std.val[0]* md->lambda/outputLambda;
+            double stdVal = std.val[0]* md->currentSettings().lambda/outputLambda;
             if (stdVal > filterRMS){
                 QFileInfo info(name);
                 QString item = QString("%1 RMS:%2").arg(info.baseName()).arg(stdVal, 0, 'f');
--- a/mirrordlg.h
+++ b/mirrordlg.h
@@ -32,40 +32,32 @@ class mirrorDlg : public QDialog
     Q_OBJECT
 
 public:
-    //TODO if we get rid of the singleton design, settings are read from settingsFacade instead here
+    // TODO if we get rid of the singleton design, settings could be read from settingsFacade instead here
     // check if it makes sense
     static mirrorDlg *get_Instance();
     ~mirrorDlg();
     mirrorDlg(const mirrorDlg&) = delete;
     mirrorDlg& operator=(const mirrorDlg&) = delete;
 
-    // File and configuration operations
-    void loadFile(QString & fileName);
-    void updateZ8();
+    // TODO this group must be investigatedto validate they are OK
+    // we cannot change settings/configuration from both outside and inside the dialog. We need to have a single source of truth for settings/configuration.
     void updateAutoInvertStatus();
-
-    // Computed/derived values (read-only, not from settings)
-    bool mm;
-    double FNumber;
-    double z8;
-    static QString m_projectPath;
-    
-    // State flags
-    bool m_obsChanged;
-    bool m_majorHorizontal;
-
-    // Methods for configuration access/modification
-    void on_roc_Changed(const double newVal);
-    void on_diameter_Changed(const double diam);
-    bool shouldFlipH();
-    static QString getProjectPath();
     void newLambda(const QString &v);
-    double getMinorAxis();
-    bool isEllipse();
     void setMinorAxis(double val);
     void setVerticalAxis(double val);
     void setOutlineShape(outlineShape shape);
-    void setObsPercent(double obs);
+    void on_roc_Changed(const double newVal);
+    void on_diameter_Changed(const double diam);
+    
+    
+    // Computed/derived value accessors (read-only)
+    double getFNumber() const { return FNumber; }
+    double getZ8() const { return z8; }
+    static QString getProjectPath();
+    double getMinorAxis();
+    bool isEllipse();
+    bool shouldFlipH();
+
     
     /** @brief Access current mirror settings.
      *  Returns the persistent copy - the last saved state.
@@ -73,11 +65,7 @@ public:
     const MirrorSettings& currentSettings() const { return m_current; }
     
 private:
-    // Persistent mirror configuration copy (source of truth for external code)
-    MirrorSettings m_current;
-    
-    // Working copy for dialog edits (discarded on Cancel, committed to m_current on OK)
-    MirrorSettings m_draft;
+
     
 private slots:
     void on_ReadBtn_clicked();
@@ -154,10 +142,30 @@ private:
     void loadDraftFromSettings();
 
     Ui::mirrorDlg *ui;
+
+    // State flags
     bool m_aperatureReductionValueChanged;
+    bool m_obsChanged;
+
     QTimer spacingChangeTimer;
+
+    // Persistent mirror configuration copy (source of truth for external code)
+    MirrorSettings m_current;
+    
+    // Working copy for dialog edits (discarded on Cancel, committed to m_current on OK)
+    MirrorSettings m_draft;
+    
+    // Computed/derived values (read-only, not from settings)
+    //TODO actually mm is not saved in settings. should probably be saved as it's a user preference. I need to check what happens when user enters values in mm or inch and reopens DFTFringe
+    bool mm;                           // Unit display flag: true = mm, false = other units
+    double FNumber;                    // Computed f-number (focal length / diameter)
+    double z8;                         // Z8 Zernike coefficient or null reference value
+    static QString m_projectPath;      // Current project directory path
+    
     void saveJson(const QString &fileName);
     void enableAnnular(bool enable);
+    void updateZ8();
+    void loadFile(QString & fileName);
     
 };
 
--- a/mirrordlg.h
+++ b/mirrordlg.h
@@ -21,6 +21,7 @@
 #include <QDialog>
 #include <QTimer>
 #include "autoinvertdlg.h"
+#include "settingsfacade.h"
 
 namespace Ui {
 class mirrorDlg;
@@ -133,15 +134,29 @@ signals:
     void recomputeZerns();
     void aperatureChanged();
 
+protected:
+    /** @brief Reload draft settings before dialog becomes visible.
+     *  Ensures Cancel always reverts to the last-saved state (issue #121). */
+    void showEvent(QShowEvent *event) override;
+
 private:
     explicit mirrorDlg(QWidget *parent = 0);
     void setclearAp();
+    
+    /** @brief Load draft from persistent settings before dialog is shown.
+     *  Ensures Cancel always reverts to the last saved state. */
+    void loadDraftFromSettings();
 
     Ui::mirrorDlg *ui;
     bool m_aperatureReductionValueChanged;
     QTimer spacingChangeTimer;
     void saveJson(const QString &fileName);
     void enableAnnular(bool enable);
+    
+    /** @brief Working copy of mirror settings during dialog edit.
+     *  All UI modifications update this draft. On OK, it persists via facade.
+     *  On Cancel, it's discarded, leaving member variables unchanged. */
+    MirrorSettings m_draft;
 };
 
 #endif // MIRRORDLG_H
--- a/mirrordlg.h
+++ b/mirrordlg.h
@@ -60,6 +60,7 @@ public:
     double getZ8() const { return z8; }
     static QString getProjectPath();
     double getMinorAxis();
+    double getClearAperture() const;
     bool isEllipse();
     bool shouldFlipH();
 
--- a/mirrordlg.h
+++ b/mirrordlg.h
@@ -39,7 +39,7 @@ public:
     mirrorDlg(const mirrorDlg&) = delete;
     mirrorDlg& operator=(const mirrorDlg&) = delete;
 
-    // TODO this group must be investigatedto validate they are OK
+    // TODO this group must be investigated to validate they are OK
     // we cannot change settings/configuration from both outside and inside the dialog. We need to have a single source of truth for settings/configuration.
     void updateAutoInvertStatus();
     void newLambda(const QString &v);
--- a/mirrordlg.h
+++ b/mirrordlg.h
@@ -117,13 +117,7 @@ private slots:
     void on_btnChangeAutoInvert_clicked();
 
 signals:
-    // TODO some of these are not used. 
-    // also notify shoud probably only happen when OK is clicked, not on every change.
-    void diameterChanged(double);
-    void rocChanged(double);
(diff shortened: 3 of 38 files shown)


astigstatsdlg.cpp
```diff
@@ -489,8 +490,7 @@ void astigStatsDlg::showItem(const QVariant &item, bool on, int /*ndx*/){
 
 void astigStatsDlg::on_zernikePB_pressed()
 {
-    QSettings set;
-    QString path = set.value("mirrorConfigFile").toString();
+    QString path = SettingsFacade::instance().appStore().load().mirrorConfigFile;
```
atsju, line 493: all direct access to Qsettings have been repalced with settingsFacade

averagewavefrontfilesdlg.cpp
```diff
@@ -66,7 +66,7 @@ void averageWaveFrontFilesDlg::on_process_clicked()
             sm->generateSurfacefromWavefront(wf);
             cv::Scalar mean,std;
             cv::meanStdDev(wf->workData,mean,std,wf->workMask);
-            double stdVal = std.val[0]* md->lambda/outputLambda;
+            double stdVal = std.val[0]* md->currentSettings().lambda/outputLambda;
```
atsju, line 69: all parameters like lambda, diameter and other are in a settings struc of the mirror for better access control and management when settings are modifed

mirrordlg.h
```diff
@@ -123,27 +123,51 @@ private slots:
...
     bool m_aperatureReductionValueChanged;
+    bool m_obsChanged;
+
     QTimer spacingChangeTimer;
+
+    // Persistent mirror configuration copy (source of truth)
+    MirrorSettings m_current;
+    
+    // Working copy for dialog edits (discarded on Cancel, committed to m_current on OK)
+    MirrorSettings m_draft;
```
atsju, line 157: mirror settings (draft during edition and current when accepted) are no private and managed with accessors.

settingsfacade.cpp
```diff
@@ -0,0 +1,53 @@
+#include "settingsfacade.h"
```
atsju, line 1: facade file is scalable and supposed to be the only entry point for every Qsettings (not in this PR, but in extremely long term it's doable if we want)

settingsstores.cpp
```diff
@@ -0,0 +1,77 @@
+#include "settingsstores.h"
+
+#include <QSettings>
+
+#define SETTINGS_STORE_LOAD_FIELD_FROM_QSETTINGS(type, name, defaultValue, key, converter) \
```
atsju, line 5: the store is accessed by facade and uses some macro to read the settingsstores_fields.inc file

settingsstores_fields.inc
```diff
@@ -0,0 +1,51 @@
+// Shared settings schema list.
+// Entry format for callbacks:
+// FIELD(type, memberName, defaultValue, key, converter)
```
atsju, line 3: this is the really neat part
`FIELD(type, memberName, defaultValue, key, converter)`
Everthing is typed here only once. It's the main source a truth.
I know Dale has been extremelly carefull to copy paste the same default aclues, key strings, converters but this solved programatically the risk of error.

It also centralises all the used keys in one knwoledge file. I did not change any key to not break compatibility.

atsju (commented):
Hi @gr5 
while this PR is not ready for review, could you have a look at it and tell me honnestly what you think about it ?
It's quite large and indirectly does more than title says.
I added some comments in the review pane and in presentation to help out.
The build should be functionnal except for elipse related features.

I have a new personal project restoring an old lathe, so I would rather not spend hours finishing this if it doesn't pass initial smell test. I already spent quite some time to arrive here.

gr5:
I will try to get to this tomorrow.  I spent most of my days away from the computer today and yesterday.  Tomorrow I should have more time.

settingsstores_fields.inc
gr5: I think this should be ".h".  I read about .inc but I use 3 different editors regularly to look at the code.  QT is great where I can right click something and find it's definition.  But the other editors I have to specify which type of file to search through and I don't want to add "*.inc" or "*.*".

You didn't define FIELD macro/function yet, right?  I didn't find that anywhere.  I assume that's what you meant by you still had work to do.

gr5:
I love the bit where you have m_draft and m_current.

By the way please add an "Apply" button to the mirror dialog so people can see the result of their change without closing the dialog.

I know your love your simplification of storing these variables but there are only 15.  Someone trying to understand the code would normally look at the 15 in the .h file and possibly get to see comments about each one explaining things like the units (mm versus inches) and other helpful details.  Not sure where to look for that info.  I guess in the ".inc" file.

I don't hate it.  I don't love it.

If Dale wants to add another member variable to mirrordlg, will he figure this out?  Will he know to go to the settingsstores_fields.inc file?

It is pretty nice and elegant.  You have such ambition - I love that part.

Let me talk to Dale.

settingsstores_fields.inc
atsju: I can rename to .h no problem. It was .inc because it's not "really" code. 

FIELD is not a macro this all already works.

search for SETTINGS_STORE_FOR_EACH_MIRROR_FIELD and SETTINGS_STORE_DECLARE_STRUCT_FIELD for example

atsju:
Yes "apply" button is a good idea.

I also had in mind to add comments about each of the qsettings into the .inc to centralise it but somehow I didn't end doing it.

Note the settings store is merely a proof of the concept and is applied only for mirror settings. there are plenty of other Qsettings that could use the same way (or not). I believe it's arround 210 settings in total [QsettingsList.txt](https://github.com/user-attachments/files/31917611/QsettingsList.txt).

Edit: I probably misunderstood you were talking about the 15 member variables in stuct or the 15 Qsettings. Anyway, most of comment is true for both.

atsju:
Thank you for having looked at it !

settingsstores_fields.inc
gr5: Maybe put a few starter comments in there.  Like how roc and diameter are in mm.  That will remind/encourage me to add other comments in the future if I'm trying to understand some setting.

In QT creator I put the cursor on a line of code in averagewavefrontfilesdlg.cpp where it was looking at lambda.  I hit F2 key and it properly jumped to correct line in the .inc file.  So if there was a comment there I would see it.

It is definitely nice to see the defaults there on the same line as the variable type (double) and what it's called in the registry.

githubdoe:
I want to make sure you know that mirror settings should never be saved in the QSettings as the source for the mirror project.  They are unique to each mirror and test setup.  So they must be read at startup each time from the current mirror configuration file.  Not from QSettings.

gr5:
What?  What if there is no mirror file?  And if the mirror is set to the latest mirror file, then qsettings should also reflect the settings of the latest mirror file anyway so why would it matter.

Isn't the way it works now is that it's stored in qsettings?

gr5:
When you say "mirror project" are you referring to this PR?  This pr is about mirrorDlg.  I assume that's what you mean by "mirror project".

gr5:
Dale I sent you a long email yesterday.  I hope you got it.

githubdoe:
No I did not get it.

githubdoe:
mirror project is the thing a user creates when testing a mirror.  Usually unique to each mirror.  Usually a directory with at least a mirror config file.  From dftfring you switch mirror projects by loading a different mirror config.  IIRC the source of truth for DFTFringe is the mirror dialog which was created from the mirror config.  I don't remember if any classes use the Qsettings to get mirror config data.  But if they do that was what I use to do.  These days I always use the mirror dialog.  It is a singleton just for that purpose.  

Yes the mirror dialog stores it state in QSettings and recovers them. from there.  Yes some people like an apply button.  I don't think I ever thought it needed one.  It would be nice to allow the user to ack out recent changes when they made a mistake.  I was just too lazy to implement that.

gr5:
> Dale I sent you a long email yesterday. I hope you got it.

I just resent it.

PR #371, opened 2026-08-12
import wft config yes button was not working at all

atsju:
#358 is not fully implemented as proposed in discussion. 
But this is a very minimal fix because the "yes" button has not any effect at all.
I need it for other PR and I don't want to do the full correction because it would create unnecesary conflicts with my own branches at this point.


gr5:
Wow.  I see that the wrong variable was compared.

gr5 (approved):
code looks good.  Tested and it works.  Works with npz files as well 😄

Issue #374, opened 2026-08-16
ignore edge region is unreliable


atsju:
More information comming later.

<img width="643" height="406" alt="Image" src="https://github.com/user-attachments/assets/412ea64c-6f84-4e79-88d6-48b40f7ca974" />

Testing with DFTFringe 8.4.0 the result of these is sometimes very strange and clearly wrong. Display impact is not always same depending on igram.

atsju:
## List of issues I identified :

min and max value for spinbox should be tied to mirror dimensions.
- outside edge can go -500 to +5; Do we actually want potitive values ?
- inside edge can go -99 to +99; Same, do we actually need  negative values ? 
- DFDFringe doesn't like when the values are more then mirror radius, it's not user/idiot proof

------------------
I think `wf->m_inside` uses default constructor somewhere when igram has no center hole. But this leads to wrong center coordinates and bad mask. Can only be reprduced with small igrams or you won't see the effect of coordinates -1/-1

<img width="1607" height="518" alt="Image" src="https://github.com/user-attachments/assets/28c2cb59-42c7-4934-9154-1a51da22fb76" />

---------------

outside edge changes shape of wavefront. As far as I understand, this is wrong/unexpected.

<img width="1534" height="927" alt="Image" src="https://github.com/user-attachments/assets/86422c48-b39c-406b-a384-462e627f5e2a" />

<img width="1563" height="953" alt="Image" src="https://github.com/user-attachments/assets/a3f4f4a5-30f5-4c75-89f2-55c058812ff9" />

atsju:
also drawn center edge is larger than what is selected in spinbox
(outside edge looks OK)

PR #375, opened 2026-08-16
Fix ignore edge region. Center edge had wrong size and could have wrong coordinates

atsju:
Somewhat improves #374. It does not fix all identified issues. What is fixed here:

- inside circle is not always defined. So we use outside circle coordinates. They are concentric anyway.
- `insideOffset` was added twice leading to wrong mask size

### DIFF
--- a/surfacemanager.cpp
+++ b/surfacemanager.cpp
@@ -641,8 +641,6 @@ void SurfaceManager::makeMask(wavefront *wf, bool useInsideCircle){
         rado += 1;
     }
 
-    double cx = wf->m_inside.m_center.x();
-    double cy = wf->m_inside.m_center.y();
     cv::Mat mask = cv::Mat::zeros(height,width,CV_8U);
     mirrorDlg &md = *mirrorDlg::get_Instance();
     double rx = radm;
@@ -668,7 +666,8 @@ void SurfaceManager::makeMask(wavefront *wf, bool useInsideCircle){
 
     if (rado > 0 && useInsideCircle) {
         uchar color = 0;
-        fillCircle(mask, cx,cy,rado, &color);
+        // inside circle is not always defined. So we use outside circle coordinates. They are concentric.
+        fillCircle(mask, xm, ym, rado, &color);
     }
 
     // expand the region by 10%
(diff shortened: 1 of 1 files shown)


gr5:
Okay I have mixed feelings here.  But in general I don't think I like this pr.  You might not know this but you can move the inner outline when you are drawing the inner and outer circles on the igram:

image

You just draw the inner outline and then use the arrows to move it around.  I don't think Dale ever intended this to work but... it works.  I spent weeks fixing bugs when you average multiple of these wavefronts where the inner circles don't line up.  Or are different diameters. Regions as well had issues with averaging but it's all fixed many versions ago.

So:
1) We could get rid of this feature where you can move the inner diameter by an offset.  I mean it's basically just another "region".  It's different from the annulus feature even though in our heads it's closely related.
2) Better: check to see if the inner region exists and is valid before using those center coordinates.
3) Best?: Disable this feature ("Ignore Edge Region - Center Edge") when there is no inner outline

I don't understand what this is supposed to fix.  If there is no inner outline, this feature doesn't seem to do anything.  And if there _is_ an inner outline it seems to work fine.

In the image below, the upper black circle is the inner outline feature and the lower black circle is from the mirror configuration dialog: "obstruction".  This is all without even getting into the annulus feature!

image

I understand that there may also be a bug related to calculating annular nulls but this PR doesn't try to address that.

atsju:
OK I see. 

You need to test 8.4.0 with a small igram (100pixels) to see the issue.
Center Edge has an effect but on coordinates -1/-1. Thus you will see it only on smaller igrams with larger center edge.
And it this case you will also see that the center edge has wrong dimension.

That being said, I confirm I didn't know the "central" hole could be moved. So I agree with your conclusions that this PR should not be merged like this. Because I specifically made assumption about the central hole being concentric. 

I will probably go with solution 2 : check validity. I don't like solution 3 as masking center always makes sense.

atsju:
I changed what we discussed about above.
I think this should be merged as it fixes some bug.

There are still overlapping cases between "obstruction" "annulus" and this but this PR doesn't change core behavior. It merely fixes display issues (used coordinates and wrong size)

gr5 (approved):
This does indeed fix a bug.

I'd love to see some tool tips for this feature and outside edge as well explaining why you can't just mess with the outer outline (because it messes up null).

PR #381, opened 2026-08-24
modified how 3D mouse buttons work

### LINKED ISSUE #380: 3D view mouse controls revamp
I have always hated the mouse controls on the 3D image.  When I had to switch to the Qt version of the 3D plotter those controls changed from my old version.  I'm constantly using the left mouse click to rotate the model and of course that does not do that.  Then you need to know the magic pressing that reverses you make to normal.

I realized the AI might know who to override that operation.  Yes it wrote the code for me.  So I'm thinking of making those controls similar to what most modern 3D programs use like Fusion360 or Blender.  

I wanted to discuss that here.  So feel free to add suggestions.  I will create a new branch from master to do that work in.  Controls?

githubdoe:
Modified what 3D graph mouse button and wheel does.  Left rotates, right pans, wheel zooms.

Closes #380

### DIFF
--- a/custom3dinputhandler.h
+++ b/custom3dinputhandler.h
@@ -1,0 +1,27 @@
+#ifndef CUSTOM3DINPUTHANDLER_H
+#define CUSTOM3DINPUTHANDLER_H
+
+#include <QtDataVisualization/Q3DInputHandler>
+#include <QtDataVisualization/QAbstract3DGraph>
+#include <QtDataVisualization/Q3DCamera>
+#include <QMouseEvent>
+
+using namespace QtDataVisualization;
+
+class Custom3DInputHandler : public Q3DInputHandler {
+    Q_OBJECT
+public:
+    Custom3DInputHandler(QAbstract3DGraph *graph = nullptr);
+
+protected:
+    void mousePressEvent(QMouseEvent *event, const QPoint &mousePos) override;
+    void mouseReleaseEvent(QMouseEvent *event, const QPoint &mousePos) override;
+    void mouseMoveEvent(QMouseEvent *event, const QPoint &mousePos) override;
+
+private:
+    bool m_isRightDragging;
+    QPoint m_lastMousePos;
+    QAbstract3DGraph *m_graphRef;
+};
+
+#endif // CUSTOM3DINPUTHANDLER_H
--- a/custom3dinputhandler.h
+++ b/custom3dinputhandler.h
@@ -6,7 +6,10 @@
 #include <QtDataVisualization/Q3DCamera>
 #include <QMouseEvent>
 
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+//keep compatiblity with Qt5
 using namespace QtDataVisualization;
+#endif
 
 class Custom3DInputHandler : public Q3DInputHandler {
     Q_OBJECT
--- a/surfacegraph.cpp
+++ b/surfacegraph.cpp
@@ -44,8 +44,6 @@ const float sampleMin = -8.0f;
 const float sampleMax = 8.0f;
 #include <QOpenGLFunctions>
 
-#include <Q3DInputHandler>
-#include <QMouseEvent>
 
 
 
--- a/surfacegraph.cpp
+++ b/surfacegraph.cpp
@@ -36,17 +36,26 @@
 #include <QLabel>
 #include <QSettings>
 #include <QDebug>
-
+#include "custom3dinputhandler.h"
 int sampleCountX = 50;
 int sampleCountZ = 50;
 
 const float sampleMin = -8.0f;
 const float sampleMax = 8.0f;
 #include <QOpenGLFunctions>
+
+#include <Q3DInputHandler>
+#include <QMouseEvent>
+
+
+
 SurfaceGraph::SurfaceGraph(Q3DSurface *surface)
     : m_graph(surface),m_axisMaxSliderXValue(100),m_axisMinSliderXValue(0), m_wf(0), m_colorRange(.56)
 {
     m_xGridMin = 0;
+    // Replace default input handler with your custom one
+    Custom3DInputHandler *customInputHandler = new Custom3DInputHandler(m_graph);
+    m_graph->setActiveInputHandler(customInputHandler);
     m_graph->setAxisX(new QValue3DAxis);
     m_graph->setAxisY(new QValue3DAxis);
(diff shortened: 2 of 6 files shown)


surfacegraph.cpp
```diff
@@ -36,17 +36,26 @@
...
+#include "custom3dinputhandler.h"
 int sampleCountX = 50;
 int sampleCountZ = 50;
 
 const float sampleMin = -8.0f;
 const float sampleMax = 8.0f;
 #include <QOpenGLFunctions>
+
+#include <Q3DInputHandler>
+#include <QMouseEvent>
```
atsju: those are not needed. they are managed by #include "custom3dinputhandler.h" and ony needed by itself.

Remove from here.
```suggestion
```

custom3dinputhandler.h
```diff
@@ -0,0 +1,27 @@
+#ifndef CUSTOM3DINPUTHANDLER_H
+#define CUSTOM3DINPUTHANDLER_H
+
+#include <QtDataVisualization/Q3DInputHandler>
+#include <QtDataVisualization/QAbstract3DGraph>
+#include <QtDataVisualization/Q3DCamera>
+#include <QMouseEvent>
+
+using namespace QtDataVisualization;
```
atsju, line 11:
```suggestion
#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
//keep compatiblity with Qt5
using namespace QtDataVisualization;
#endif
```

atsju:
Hi Dale, 
Congrats on the git usage. Looks like you are mastering it.

I fixed automated build. There are some remaining deprecation warnings that need to be fixed but the build is OK.

The feature is absolutely amazing and I love it.

Being used to 3D sodftwares, there is one flaw though:
right click + up/down or left/right motion always moves allong Y and X repectively. This is only good when you look from top.
If you ever rotated the 3D, all movements become more or less nonsensical.

Here is a capture of the "problem". Video was too large so there are 2 parts.

https://github.com/user-attachments/assets/fb927c69-8c56-433d-a345-d49c009ad9e4
https://github.com/user-attachments/assets/acf36f8c-24e5-4996-b945-d96e075e2ea5


But to be honnest I like this that much that I would like to get it merged and enhanced in a different PR.

One last note: I would advocate to make the right click "orbit". Because this is the way it was in the past. And while I don't care  which click does what, I think it's better to keep some some of backcompatiblity and only modify/enhance what left click is doing.

githubdoe:
The older versions use to be left click orbit.  So that just restored to the old way and the way I find much 3D apps that I use.

githubdoe:
Yes pan gets a little weird if you rotate too much.  I bet we could get AI to fix that by remembering the current rotation.

githubdoe:
Currently have the pan fixed in my local temp branch.   Now trying on my local master.

atsju:
Dale, do no forget to `git pull` your branch before you start modifications. That'"s because I made 2 commits on your remote branch for the .pro files and QT5 compatibility.

I do not expect any conflicts. But it's easier it you pull from the start than doing after you modified/commited on your local copy

githubdoe:
Yes I finally got the pull figured out and the push is now done.

atsju:
Nice. 
I will have less time for the rest of the day.  George will proably try it out too.

If you could fix the deprecation warnings too it would be great. I'm unsure if Qt5 supports `position()` and `globalPosition()` so you might need an ifdef.

Thank you !

atsju:
I tested it. It's much better !

I belive mouse click + Y moving is not working on the correct axis. It does the same as scrolling with weel when in fact it shall behave similarly as X moving.

I might not be 100% clear. Let me know and I can do a video

githubdoe:
Added the ifdef for QT5 and Qt6.  I wish I had the time to switch to Qt6.  I don't like those flags.  I know George once showed my how to switch but that has been lost by me.  

I think the mouse does what I wanted it to do.  Which is show the wave front from any reasonable angle zoom it and pan it.  i was surprised when I opened Fusion 360 that orbit is on the right button and my muscle memory took over and used it to orbit.  Yet when I use DFTFringe that muscle memory is for the left button because that was what it was for years.   So it will mess with those that never knew left orbit for DFtfringe.  Sorry.

gr5:
This removes nothing - it only adds more capability.  Switching between qt5 and qt6 doesn't even require exiting qt creator after you follow these directions.

https://youtu.be/ZHo5vspf6BQ

Unfortunately the standard project files have all relative paths so if you have many different cloned repositories, you have to add the new folders to each repository.  I suppose that only takes a few more minutes.

gr5 (approved):
Tried it out.  Didn't build it as Dale already did.  This feature works great.  Code looks great.

Dale this can be merged now.  You want to click the button here on github or is it okay if I do it?  Do you have more changes in mind or is this ready in your opinion.  It is ready in my opinion.

githubdoe:
It is ready to merge.  Go ahead.

atsju:
Hi I don't know fusion 360 and I know each softwawre has different ways of working.

So just confirm you actually wanted to implement truck/dolly which is unusual for me. but again I don't know fusion 360.


https://github.com/user-attachments/assets/cea48d66-eae6-42de-b258-71436b57b5cd



Because I expected panX/panY



https://github.com/user-attachments/assets/1aa0c660-a326-4ff8-840c-62d9af67e0ba

githubdoe:
I just want some way to move what is off screen back into full view.  It seems to do that.  I don't want to spend a lot of time fighting the math involved moving the camera correctly which is what it is trying to do.  So for not I'm happy with it.  Before if the edge decided to go off screen it was a hassle to zoom to get it back.

gr5:
I tried it in the overhead orientation and it seemed fine but yeah that could be fixed by changing the equation slightly.  I don't know if you want to fix it @atsju or I could fix it.

We already have the mouse scroll for zooming in and out and everything is great when are not in "overhead" mode.  But yeah in overhead view it's a little wonky.

gr5:
Presumably people who primarily want overhead view will be using the contour map instead.  Which also allows you to zoom in.

atsju:
I believe it's not only in overhead. It's just more obvious in any orientation where you face a plane 90°.

It's not wrong per se. It's just unusual way of working for a 3Dview and as you say, we already have mouse wheel for zoom.

I let you modify it if you have time. I should take some time working on my other PR in progress.

gr5:
Let me fix this.

gr5:
see pr #382

PR #382, opened 2026-08-26
3d view: this fix makes it so that right drag pans much better when moving mouse up/down

gr5:
title says it all.

### DIFF
--- a/custom3dinputhandler.cpp
+++ b/custom3dinputhandler.cpp
@@ -72,22 +72,22 @@ void Custom3DInputHandler::mouseMoveEvent(QMouseEvent *event, const QPoint &mous
             float xRotRad = qDegreesToRadians(camera->xRotation()); // Yaw (horizontal orbit)
             float yRotRad = qDegreesToRadians(camera->yRotation()); // Pitch (vertical tilt)
 
-            float cosX = qCos(xRotRad);
-            float sinX = qSin(xRotRad);
-            float cosY = qCos(yRotRad);
-            float sinY = qSin(yRotRad);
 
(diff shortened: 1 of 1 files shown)


atsju:
I have seen your comment, build can be a bit long, especially on a new PR when cache needs to be rebuilt.

I find this version much more natural. Still slightly strange on Y when the 3D is at 45°. It still moves forward and backward slighty.
But it's good to go for me.

atsju (approved):

gr5:
@atsju and @Lucassifoni  - it looks like this pr broke something related to the mac release?  It didn't fail until after I did the merge for some reason.

Lucassifoni:
I’ll have a look tomorrow, it seems that the build went fine on both architectures but the merging to a single binary failed. Also it looks like subsequent action runs on your new release preparation passed fine ? It might be a flake, GitHub is pretty unreliable lately.In any way I wanted to say that you can totally ping me re: the Mac version, I’m eager to help maintain its existence over time Le 26 août 2026 à 21:43, gr5 ***@***.***> a écrit :﻿gr5 left a comment (githubdoe/DFTFringe#382)
@atsju and @Lucassifoni  - it looks like this pr broke something related to the mac release?  It didn't fail until after I did the merge for some reason.

Lucassifoni:
@gr5 I don't find a plausible cause for this failing build, especially since the next runs did not fail. Could you re-run the failing workflow to determine if this was a flake on github's end ? I find the error message of differing files between the two executables surprising.

gr5:
I think we should just ignore that error.  It was probably a problem with github.  It's building fine now.

PR #384, opened 2026-08-27
Universal builds macos fixes

Lucassifoni:
The mac OS build chain is fine, but the combination of the two packages into a single one had logic that verified both contained the same artifacts.. which can be wrong if the action runners package different versions of libraries, which will periodically happen as arm drifts further away from intel macs.

Here it was ffmpeg pulled by openCV that differed. The universal build reconcilation job is simpler now.


Lucassifoni:
It passed : https://github.com/githubdoe/DFTFringe/actions/runs/33109740800/job/98655654133

gr5 (approved):

